PR #1, opened 2018-10-27
Dr. Scannell's first cut

kscanne:
Please review my janky Perl code!


kscanne:
NB: reviewer can run all test cases as follows:
$ cd ks
$ perl strip.pl

PR #6, opened 2018-10-30
added .class and test files

waqarmuhammad1:


wm/README
```diff
@@ -1 +1,167 @@
...
+import java.util.logging.Logger;
+public class Assignment {
+
+    /**
+     * @param args the command line arguments
+     */
+    public static void main(String[] args) {
+        try {
+            // TODO code application logic here
+            System.out.println(parse("/home/waqar/Git-Repo/5030/tests/cases.tsv"));
```
seandough3rty: Static path, maybe try relative?

wm/README
```diff
@@ -1 +1,167 @@
...
+    public static ArrayList<String> parse(String file_path) throws FileNotFoundException, IOException, Exception{
+    
+        ArrayList<String> data_list = read_file(file_path);
+        ArrayList<String> filtered_data = new ArrayList<String>();
+        for(String dat: data_list){
+            
+            String[] entry = dat.split("\t");
+            
+            String lang_symbol = entry[0];
+            String[] data = entry[1].split(" ");
```
seandough3rty: Could cause filtering to fail if there are two spaces next to each other.

waqarmuhammad1: Space is considered as a single character no matter how many consecutive spaces there are it will work fine.

PR #9, opened 2018-10-30
stripStopwords implemented

MatteoPE:
This is my pull request


MatteoPE:
After the last commit, all tests pass, except for
line 73: el    totally valid text is here    Error: Not a valid language code
because it is not consistent with 
line 64: lala    this is non-existing language test case and expects no stripping    this is non-existing language test case and expects no stripping

mf/StripStopwords/src/com/company/Main.java
```diff
@@ -5,14 +5,24 @@
...
+ */
 public class Main {
 
     public static void main(String[] args) {
-        // write your code here
+
         try {
-            String str = "a la mode";
-            Stripper s = new Stripper("//home//matteo//SoftwareDev//5030//data");
-            System.out.println(s.stripStopwords("fr", str));
```
nyrabbit: nice work on cleaning unnecessary comment!

MatteoPE: thank you

PR #12, opened 2018-10-30
Jyckle version

Jyckle:
This one should work, please review!


jp/stripStopWords.js
```diff
@@ -0,0 +1,26 @@
+module.exports={
+	strip_stopwords: function(languageCode, inputText){
+		var fileSystem = require('fs');
+		var validLanguageCodes = fileSystem.readFileSync('../data/list.txt').toString().split('\n');
+		var languageCodeValid = validLanguageCodes.indexOf(languageCode);
+
+		if (languageCodeValid === -1){
+			return "Error: Not a valid language code"
```
pannapat: We may need to agree a consistent error message for invalid language case.

jp/stripStopWords.js
```diff
@@ -0,0 +1,26 @@
...
+		}
+
+		var stopWords = fileSystem.readFileSync('../data/' + languageCode + '.txt').toString().split('\n');
+		var splitInput = inputText.toString().split(" ");
+		var finalString= "";
+
+		for (word in splitInput){
```
pannapat: This is very minor stuff. Due to the fact that JavaScript has a strange for-in looping, we could make this 'word' variable clearer by naming it as 'index' or 'i'.

jp/testStripStopWords.js
```diff
@@ -0,0 +1,35 @@
...
+	console.log("Tests Passed: " + passedTests);
+	console.log("Tests Failed: " + failedTests);
+}
+
+function runTest(testCase){
+	var output = stripFunction.strip_stopwords(testCase[0],testCase[1]);
+	if (output !== testCase[2]){
+		console.log("Error: should have returned: " + testCase[2]);
+		console.log("But instead returned: " + output);
+		return "Fail";
```
pannapat: It may be easier to debug when we have a lot of test cases if we also print out the test case number or something that help us traceback to the failed test case.

jp/README
```diff
@@ -1 +1,17 @@
...
+
+The function can be used by importing the stripStopWords.js file as follows:
+
+var stripFunction = require('./stripStopWords');
+
+Then, the function itself may be called as:
+
+stripFunction.strip_stopwords(languageCode,inputText);
+
+To run the tests for this function, simply run:
```
pannapat: This is very useful to document execute steps in README. Thank you.

jp/stripStopWords.js
```diff
@@ -0,0 +1,26 @@
...
+	strip_stopwords: function(languageCode, inputText){
+		var fileSystem = require('fs');
+		var validLanguageCodes = fileSystem.readFileSync('../data/list.txt').toString().split('\n');
+		var languageCodeValid = validLanguageCodes.indexOf(languageCode);
+
+		if (languageCodeValid === -1){
+			return "Error: Not a valid language code"
+		}
+
+		var stopWords = fileSystem.readFileSync('../data/' + languageCode + '.txt').toString().split('\n');
```
pannapat: It could be an undefined exception if the specified language's stop words file does not exist. For example, there's 'kr' language in data/list.txt  but the data/kr.txt does not exist.

jp/stripStopWords.js
Jyckle: Agreed, That would definitely help clean things up!

jp/stripStopWords.js
Jyckle: Great point! I will adjust that to make it more clear!

jp/testStripStopWords.js
Jyckle: I completely agree, I actually loved seeing that in your code, so I went in and adjusted in mine! Makes it way easier for debugging

jp/stripStopWords.js
Jyckle: That possibility didn't even cross my mind until you mentioned it! Great catch! I will add code to handle that case

PR #16, opened 2018-10-31
Stripper version2

Matthew-Mosior:


mm/stopwordstripper.py
```diff
@@ -0,0 +1,114 @@
...
+    #Define stopdict (for loading in stop word dictionaries).
+    stopdict = []
+    largestopdict = []
+    #Define stripped (checkline after going through selective stripping).
+    strippedlist = []
+    #Define englishlist final list.
+    englishlist = []
+    #Open the stop dictionary file based on langcode.
+    if (langcode == "en"):
+        with open("en.txt","r") as file:
```
cletusajibade: 1) This means "en.txt" exists in the same directory ("mm") as this source code. I think we should use relative path **('../data/en.txt')** to the data directory so as to maintain one source of data,  that way when more languages or stop words are added in the future, the code will not break.

2) Another reason is we are duplicating data sources if we need to copy "en.txt" to "mm" directory. I think a good practice would be for us to have program data in one place for maintainability. Hence the relative path to the data directory.

mm/stopwordstripper.py
```diff
@@ -0,0 +1,114 @@
...
+            for line in file:
+                line = line.strip()
+                stopdict.append(line)
+    elif (langcode == "fr"):
+        with open("fr.txt","r") as file:
```
cletusajibade: Same reason given on line 54.

mm/stopwordstripper.py
```diff
@@ -0,0 +1,114 @@
...
+            for line in file:
+                line = line.strip()
+                stopdict.append(line)
+    elif (langcode == "ga"):
+        with open("ga.txt","r") as file:
```
cletusajibade: Same reason given on line 54.

mm/stopwordstripper.py
```diff
@@ -0,0 +1,114 @@
...
+            for line in file: 
+                line = line.strip()
+                stopdict.append(line)
+    elif (langcode == "gd"):
+        with open("gd.txt","r") as file:
```
cletusajibade: Same reason given on line 54.

mm/stopwordstripper.py
```diff
@@ -0,0 +1,114 @@
...
+            for line in file:
+                line = line.strip()
+                stopdict.append(line)
+    elif (langcode == "gv"):
+        with open("gv.txt","r") as file:
```
cletusajibade: Same reason given on line 54.

mm/stopwordstripper.py
```diff
@@ -0,0 +1,114 @@
...
+            for line in file:
+                line = line.strip()
+                stopdict.append(line)
+    elif (langcode == "ru"):
+        with open("ru.txt","r") as file:
```
cletusajibade: Same reason given on line 54.

mm/stopwordstripper.py
```diff
@@ -0,0 +1,114 @@
...
+        for row in tsvin:
+            strippedline = stopwords(row[0],row[1])     
+            tsvout.writerow([row[0],row[1],strippedline])
+            #Use assert to check results.
+            assert strippedline == row[2]
+
+#Function to define path of cases.tsv.
+def relpath():
+    scriptdir = os.path.dirname(__file__) 
+    relpath = "path/to/test/cases.tsv"
```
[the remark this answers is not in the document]
Matthew-Mosior: Hi Cletus! Thanks for the idea I changed the code to take a relative path.  (I didn't supply the actual path on my machine though).

mm/stopwordstripper.py
Matthew-Mosior: Thanks for the input, I will take a look at this.

PR #17, opened 2018-11-01
Yifei strip program 2.0

nyrabbit:
Review this


nyrabbit:
Some improvements on previous version

yw/Software/src/spellchecker/Read.java
```diff
@@ -0,0 +1,44 @@
...
+		  for (File f : files) {
+			  if(f.isFile()) {
+				  BufferedReader inputStream = null;
+				  
+				  try {
+					  inputStream = new BufferedReader(new FileReader(f));
+					  String line;
+					  
+					  while ((line = inputStream.readLine())!=null) {
+						  myList.add(line);
```
MatteoPE: This line is adding all the stopwords of different languages to the same list and a non-stopword may be stripped only because it is spelled the same as a stopword of another language. Should we use a different data structure that divides stopwords of different languages (like a map or an array of lists)?

yw/Software/src/spellchecker/Checker1.java
```diff
@@ -0,0 +1,162 @@
...
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class Checker1 {
```
MatteoPE: It seems that this class is not used. Shall we delete it?

yw/Software/src/spellchecker/Checker2.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Checker2 {
+
+	public String check(String arr) throws IOException {
+		Read r = new Read();
+		String[] sentence = arr.split("\\s+"); 
+		List<String> suspects = r.read();  // receive suspects
```
MatteoPE: This variable should be renamed "stopwords"

yw/Software/test/Checker1Test.java
```diff
@@ -0,0 +1,138 @@
...
+		};
+		String [] line = null;
+		BufferedReader TSVFile;
+		   //use BufferReader to read the tsv file
+		TSVFile = new BufferedReader(new FileReader("C:\\Users\\wangyifei\\software\\5030\\tests\\cases.tsv"));
+		String dataRow = "";
+		System.out.println("file read successful");
+		Checker2 c = new Checker2();
+		int i = 0;
+		String lan = null;
```
MatteoPE: This initialization is redundant as the next ones

yw/Software/test/Checker1Test.java
```diff
@@ -0,0 +1,138 @@
+import static org.junit.Assert.*;
```
MatteoPE: Should we delete libraries that are not used?

yw/Software/src/spellchecker/ReadTest.java
```diff
@@ -0,0 +1,26 @@
+package spellchecker;
```
MatteoPE: Should this variable be in test package?

yw/Software/src/spellchecker/Checker2.java
```diff
@@ -0,0 +1,32 @@
...
+		List<String> arrList = new ArrayList();
+//		System.out.println(suspects);
+	
+		for(int i = 0; i < sentence.length;i++) {
```
MatteoPE: A foreach loop could be a more elegant solution

yw/Software/test/Checker1Test.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.List;
+import java.util.StringTokenizer;
+
+import org.junit.Test;
+
+import spellchecker.Checker1;
+import spellchecker.Checker2;
+import spellchecker.Read;
+ 
+public class Checker1Test {
```
MatteoPE: I don't know why, but when I try to run this test class, I get a NoClassDefFoundError. Are your tests running on your local?

yw/Software/src/spellchecker/Read.java
nyrabbit: This is a good suggestion, I will fix this problem as soon as possible

yw/Software/src/spellchecker/Checker2.java
nyrabbit: Thanks for pointing out, the problem has been fixed

yw/Software/test/Checker1Test.java
nyrabbit: Thank you for indicating this problem, it has been fixed

yw/Software/test/Checker1Test.java
nyrabbit: All libraries that are not used have been deleted

yw/Software/src/spellchecker/ReadTest.java
nyrabbit: The reason of importing spellchecker is to use the function in checker 2, I think it is reasonable to import spellchecker

MatteoPE: Sorry, I mistyped: 
Should this class (ReadTest) be in test package?

MatteoPE:
Great work! Thank you for solving these little problems. I'm trying to understand why tests don't run on my PC, then I will approve your pull request if everything it's fine!

PR #22, opened 2021-03-06
5030-word-lower

jackzhenguo:
5030 class for word lowering task by Zhen Guo


S21/jackzhenguo/wordlower.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        self._special_strategies.append(new_strategy)
+
+
+if __name__ == "__main__":
+    # this is test code
+    wl = WordLower()
+    # read tsv file one line after line
+    all_succeed = True
+    with open('tests.tsv', 'r') as f:
```
Garrett-Reed, line 71: It is helpful to specify an encoding as an argument in open(). Since this program deals with different characters, it is reasonable to assume not all characters will be recognized correctly on all machines.

S21/jackzhenguo/wordlower.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+         It's mainly used to extend new and special strategy
+        :param new_strategy: input word we want to lower
+        """
+        self._special_strategies.append(new_strategy)
+
+
+if __name__ == "__main__":
+    # this is test code
+    wl = WordLower()
```
Garrett-Reed, line 68: We might want to create tests as their own files in the future. If we do that, we can make updates to the program code and have a higher level of confidence that the test code hasn't changed. This allows us to identify issues/changes in scope more quickly.

S21/jackzhenguo/wordlower.py
```diff
@@ -0,0 +1,79 @@
+# This module mainly solves lowering a word and also can process some special
```
Garrett-Reed, line 1: Very neat and tidy code, I like it!

S21/jackzhenguo/wordlower.py
jackzhenguo: Yes, Garrett. It's better to define a test function with test filename variable. Thank you.

PR #23, opened 2021-03-10
 Adding first time

FaizanSyed95:


S21/FaizanSyed95/Language_test.py
```diff
@@ -0,0 +1,64 @@
+import csv 
+
+Vowels=['A','E','I','O','U','Á','É','Í','Ó','Ú']
+GreekCapital=['Π','Ό','Λ','Η','Σ']
+GreekSmall=['π','ό','λ','η','ς']
+
```
hvkone, line 6: Did you find a way to implement the other characters? For example, Chinese characters.

S21/FaizanSyed95/Language_test.py
```diff
@@ -0,0 +1,60 @@
+import csv 
+
+Vowels=['A','E','I','O','U','Á','É','Í','Ó','Ú']
+Greek=['Π','Ό','Λ','Η','Σ']
+Greek_again=['π','ό','λ','η','ς']
+
+
+def Word_check(lang , word):
+    LG=list(lang)
+
```
hvkone, line 10: Maybe there's a specific reason that the language code string is split into two conditionals, but I think they could be combined into one. For example, use 'ga', instead of 'g' and 'a'.

S21/FaizanSyed95/Language_test.py
```diff
@@ -0,0 +1,60 @@
...
+        for i in range(len(wordArray)):
+            for j in range(len(Greek)):
+                if wordArray[i] == Greek[j]:
+                    wordArray[i] = Greek_again[j]
+                    
+        return ''.join(wordArray)
+        
+
+
+
```
hvkone: Instead of using the tests.tsv file directly, it may be better to use the input function for word and the language. Your function is nicely written, it would be able to decode the individual characters of the input.

S21/FaizanSyed95/Language_test.py
```diff
@@ -0,0 +1,60 @@
...
+file = open("C:/Users/Phaixan/Desktop/5030/S21/FaizanSyed95/tests.tsv",  encoding = "utf8")
+read_tsv = csv.reader(file,delimiter="\t") #auto test by reading from the file
+for row in read_tsv: # going through each row of tsv file
+    W = Word_check(row[1],row[0])
+    #print(W , row[2])
+    if W == row[2]:
```
hvkone, line 56: It would be beneficial to output the row information, if you give it a .tsv file. Just to keep track, if the input ever fails the conditionals of Word_check.

hvkone (commented):
It looks great overall.

S21/FaizanSyed95/Language_test.py
FaizanSyed95: Chinese, Thai, and Japanese do not have any lower case or upper cases the output for them is the same. So it is easier to just output whatever input you give in it. It is shown below in the function.

S21/FaizanSyed95/Language_test.py
FaizanSyed95: The reason was to save the code from different errors. Checking each string/char separately has a low risk of bugs in the code. Also, if there are other characters after this it will only detect the 1st two to give an output.

S21/FaizanSyed95/Language_test.py
hvkone: Okay, that makes sense. I was just wondering why you were calling vowels as a variable.

S21/FaizanSyed95/Language_test.py
FaizanSyed95: The reason for using the test file is, if we have many inputs for our test, it would be easier to write them all in one file, just like the professor gave us, and run them and check if these outputs match or not. My code checks all the list of inputs of word and languages and compare it with the output of the test file and gives the result.

S21/FaizanSyed95/Language_test.py
FaizanSyed95: It gives the output of all the rows, that keeps track of all the inputs and output and test it to match the test file. If possible, if we can run the code you would see the output if it matches the test file or not.

S21/FaizanSyed95/Language_test.py
hvkone: Language codes (ISO 639-1) are specific to each individual language. I understand your reason. I am just suggesting using two characters as one string.

S21/FaizanSyed95/Language_test.py
FaizanSyed95: I make that variable name to make it easier for the reviewer to understand that this variable is only be used in the Irish language. That was written in the description of this project.

S21/FaizanSyed95/Language_test.py
FaizanSyed95: Yes, they are, however, I don't think it would matter to use them as a string or separate characters. However, thank you for suggesting that I will look that up in the future, it might be less time-consuming.

PR #24, opened 2021-03-11
Adding the code

myadlapalli:


S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+        return result
+
+    else:
+        result = word.lower()
+
+    return result.lower()
+
+
+if __name__ == "__main__":
+    print("Enter the language tag")
```
harisalam, line 65: It would be better if we read the input from the test cases as per the requirement.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+    language = input()
+    print("Enter the sentence")
+    sentence = input()
+    print("Language tag: " + language + "\nSentence: " + sentence)
+    print("The converted sentence is:\n")
```
harisalam, line 70: Here we can make a change of inputting the data from the test file that was present in the git repository, and we can check the validity of the returned output.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+    print(toLower(sentence, language))
```
harisalam, line 71: At the end of the code where we receive the output from the function we have to check whether all the test cases have been met or not.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
+NO_CHANGE_LANGS = ['zh', 'ja', 'th']
```
harisalam, line 1: We can declare these variables inside the function for better reliability, otherwise, they can cause harm if not taken into consideration like its value changes without knowing

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+TURKISH_LANGS = ['tr', 'az']
+
+IRISH = 'ga'
+IRISH_VOWELS = ['A', 'E', 'I', 'O', 'U', 'Á', 'É', 'Í', 'Ó', 'Ú']
+
+GREEK = 'el'
+
+
+def toLower(sentence, lang):
+    sentence = sentence.strip()
```
harisalam, line 11: I think we went a bit out of scope, the problem was to read a word along with the language and then perform some lower operation to meet the rules described in the language rules.

harisalam: It would be better if we not hardcode the input.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+    result = ''
+    lang = lang.split('-')[0]
+
+    # Languages with no change
+    if lang in NO_CHANGE_LANGS:
+        return sentence
+
+    words = sentence.split(" ")
```
harisalam, line 19: We can follow the google style guide for better code writing principles by not adding blank lines between the code except function or class declaration.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+
+    for word in words:
+        result += ' ' + wordToLower(word, lang)
+
+    return result.strip()
+
+def wordToLower(word, lang):
+    result = ''
+    # Languages with no change
+    if lang in NO_CHANGE_LANGS:
```
harisalam, line 30: We have already performed this condition, here it can just increase the complexity of the code.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+        return word
+
+    # Irish
+    if lang == IRISH:
+        if len(word) > 1 and word[0] in ('t', 'n') and word[1] in IRISH_VOWELS:
```
harisalam, line 35: It would be better if we enclose multiple conditions in parenthesis for better understanding.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+            result = word[0] + '-' + word[1:].lower()
+        else:
+            result = word.lower()
+        return result
+
+    # Greek
+    elif lang == GREEK:
+        if word[-1] == 'Σ':
+            result = word[:-1].lower() + str(u"\u03C2")
```
harisalam, line 44: It would be better if we include the documentation of Unicode strings that is used here for guidance.

S21/myadlapalli/Language.py
```diff
@@ -0,0 +1,72 @@
...
+    elif lang in TURKISH_LANGS:
+        for ch in word:
+            if ch == 'I':
+                result += str(u"\u0131")
+            else:
+                result += ch.lower()
+        return result
+
+    else:
+        result = word.lower()
```
harisalam, line 59: I think it not required to lower the word after performing a lowering operation in the conditional blocks.

harisalam: we can write one more generalized or else block the program performs conversion if none of the languages is present in the test cases.

S21/myadlapalli/Language.py
harisalam: we can use the pandas library to read the test cases and then perform a further operations on that data.
import pandas as pd
data = pd.read_csv(file path, sep, encoding)

harisalam (changes requested):
We have to make all the following necessary.

S21/myadlapalli/Language.py
```diff
@@ -62,11 +62,13 @@ def wordToLower(word, lang):
...
-
+    r=open("tests.tsv",encoding="utf-8")
+    for line in r.readlines():
+        word,lang,exp=line.split('\t')
+        exp=exp.strip("\n")
+        actual=toLower(word,lang)
+        if exp==actual:
+            print("passed expected: {0} actual: {1}".format(exp,actual))
+        else:
+            print(repr("failed expected: {0} actual: {1}".format(exp,actual)))
```
harisalam, line 73: We can also introduce another conditional statement, which will return us the result when all the conditions have been met.

PR #26, opened 2021-03-11
Rajmain

rkarmuri:
Code review assignment files

### DIFF
--- a/S21/rkarmuri/ConvertToLowerCase.java
+++ b/S21/rkarmuri/ConvertToLowerCase.java
@@ -1,0 +1,98 @@
+//Method to convert given string to lowercase
+package com.company;
+
+import java.util.Locale;//Package to accept multiple languages
+
+public class ConvertToLowerCase {
+    public String convert(String wordSample, String language) {
+        System.out.println("Input word: " + wordSample);
+        System.out.println("Input language tag: " + language);
+        String wordLowerCase = wordSample.toLowerCase();//Conversion in English
+        String trazString = wordSample.toLowerCase(Locale.forLanguageTag("az"));//Conversion in Turkish and Azerbaijani
+        String russString = wordSample.toLowerCase(Locale.forLanguageTag("ru-Latn"));//Conversion in Russian
+        String hungarianString = wordSample.toLowerCase(Locale.forLanguageTag("hu"));//Conversion in Hungarian
+        //Conversion cases for Greek
+        String greekString = wordLowerCase.replace('Σ', 'σ').replace('Ε', 'ε').
+                replace('Α', 'α').replace('Β', 'β').
+                replace('Γ', 'γ').replace('Π', 'π').
+                replace('Ό', 'ό').replace('Λ', 'λ').
+                replace('Η', 'η').replace('Δ', 'δ').
+                replace('Ζ', 'ζ').replace('Θ', 'θ').
+                replace('Μ', 'μ').replace('Ν', 'ν').
+                replace('Ξ', 'ξ').replace('Ρ', 'ρ').
+                replace('Ω', 'ω').replace('Ψ', 'ψ').
+                replace('Χ', 'χ').replace('Φ', 'φ').
+                replace('Υ', 'υ').replace('Τ', 'τ');
+        String outputString;//Variable that is used to return the desired lowercase to test code method
+        switch (language) {
+            //This case is to convert English, English(US), English(Ireland), English(Latin) and Spanish words
+            case "en":
+            case "en-US":
+            case "en-IE":
+            case "en-Latn":
+            case "es":
+                System.out.println("String after converting to lower case: " + wordLowerCase);
+                outputString = wordLowerCase;
+                break;
+            //This case is to convert Turkish and Azerbaijan words
+            case "tr":
+            case "az":
+                System.out.println("String after converting to lower case: " + trazString);
+                outputString = trazString;
+                break;
+            //This case is to convert all Greek words
+            case "el":
+                if (wordSample.endsWith("Σ")) {
+                    //Replacing Word ending with Sigma condition
+                    String newString = wordLowerCase.replace('Σ', 'ς');
+                    System.out.println("String after converting to lower case: " + newString);
+                    outputString = newString;
+                } else {
+                    //Print and return the output lowercase in greek
+                    System.out.println("String after converting to lower case:" + greekString);
+                    outputString = greekString;
+                }
+                break;
+            //This case is to convert Russian words
+            case "ru-Latn":
+                System.out.println("String after converting to lower case: " + russString);
+                outputString = russString;
+                break;
+            //This case is to convert Hungarian words
+            case "hu":
+                System.out.println("String after converting to lower case: " + hungarianString);
+                outputString = hungarianString;
+                break;
+            //This case is to convert all Irish words
+            case "ga":
+            case "ga-IE":
+                char c = wordSample.charAt(1);
+                if (wordSample.startsWith("n") || wordSample.startsWith("t")) {
+                    if (c == 'A' || c == 'E' || c == 'I' || c == 'O' || c == 'U' || c == 'Á' || c == 'É' || c == 'Í' || c == 'Ó' || c == 'Ú') {
+                        //Adding hyphen when condition satisfies
+                        String IrishString1 = wordLowerCase.charAt(0) + "-" + wordLowerCase.substring(1);
+                        System.out.println("String after converting to lower case: " + IrishString1);
+                        outputString = IrishString1;
+                    } else {
+                        //Else for letters starting with "n" and "t", but not followed by capital vowels
+                        System.out.println("String after converting to lower case: " + wordLowerCase);
+                        outputString = wordLowerCase;
+                    }
+                } else {
+                    //Else for letters not starting with "n" and "t"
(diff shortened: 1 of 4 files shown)


S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Locale;//Package to accept multiple languages
+
+public class ConvertToLowerCase {
+    public String convert(String wordSample, String language) {
+        System.out.println("Input word: " + wordSample);
+        System.out.println("Input language tag: " + language);
+        String wordLowerCase = wordSample.toLowerCase();//Conversion in English
+        String trazString = wordSample.toLowerCase(Locale.forLanguageTag("az"));//Conversion in Turkish and Azerbaijani
+        String russString = wordSample.toLowerCase(Locale.forLanguageTag("ru-Latn"));//Conversion in Russian
+        String hungarianString = wordSample.toLowerCase(Locale.forLanguageTag("hu"));//Conversion in Hungarian
```
rebelDi, line 13: Can we combine all this variables? For example:

String worlLowerCase = wordSample.toLowerCase(Locale.forLanguageTag(language));

S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+                replace('Α', 'α').replace('Β', 'β').
+                replace('Γ', 'γ').replace('Π', 'π').
+                replace('Ό', 'ό').replace('Λ', 'λ').
+                replace('Η', 'η').replace('Δ', 'δ').
+                replace('Ζ', 'ζ').replace('Θ', 'θ').
+                replace('Μ', 'μ').replace('Ν', 'ν').
+                replace('Ξ', 'ξ').replace('Ρ', 'ρ').
+                replace('Ω', 'ω').replace('Ψ', 'ψ').
+                replace('Χ', 'χ').replace('Φ', 'φ').
+                replace('Υ', 'υ').replace('Τ', 'τ');
```
rebelDi, line 25: Do you think we can combine the letters and their replacement into arrays or list, so it would be more readable?

S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+        String outputString;//Variable that is used to return the desired lowercase to test code method
+        switch (language) {
+            //This case is to convert English, English(US), English(Ireland), English(Latin) and Spanish words
+            case "en":
+            case "en-US":
+            case "en-IE":
+            case "en-Latn":
```
rebelDi, line 32: Can we combine these cases somehow?

S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+    public String convert(String wordSample, String language) {
+        System.out.println("Input word: " + wordSample);
+        System.out.println("Input language tag: " + language);
+        String wordLowerCase = wordSample.toLowerCase();//Conversion in English
+        String trazString = wordSample.toLowerCase(Locale.forLanguageTag("az"));//Conversion in Turkish and Azerbaijani
+        String russString = wordSample.toLowerCase(Locale.forLanguageTag("ru-Latn"));//Conversion in Russian
+        String hungarianString = wordSample.toLowerCase(Locale.forLanguageTag("hu"));//Conversion in Hungarian
+        //Conversion cases for Greek
+        String greekString = wordLowerCase.replace('Σ', 'σ').replace('Ε', 'ε').
+                replace('Α', 'α').replace('Β', 'β').
```
rebelDi, line 16: Can we have a dot at the beginning of the line instead of the end like this? :
.replace('Α', 'α').replace('Β', 'β')

S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+        String outputString;//Variable that is used to return the desired lowercase to test code method
+        switch (language) {
+            //This case is to convert English, English(US), English(Ireland), English(Latin) and Spanish words
+            case "en":
+            case "en-US":
+            case "en-IE":
+            case "en-Latn":
+            case "es":
+                System.out.println("String after converting to lower case: " + wordLowerCase);
+                outputString = wordLowerCase;
```
rebelDi, line 35: Instead of printing the result at every case can we just assign the correct result to the outputString in the cases and after all the cases were checked just print the string?

S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+package com.company;
+
+import java.util.Locale;//Package to accept multiple languages
+
+public class ConvertToLowerCase {
+    public String convert(String wordSample, String language) {
+        System.out.println("Input word: " + wordSample);
+        System.out.println("Input language tag: " + language);
+        String wordLowerCase = wordSample.toLowerCase();//Conversion in English
+        String trazString = wordSample.toLowerCase(Locale.forLanguageTag("az"));//Conversion in Turkish and Azerbaijani
```
rebelDi, line 11: Can we check the code with Checkstyle for the whitespaces errors?

S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+            case "hu":
+                System.out.println("String after converting to lower case: " + hungarianString);
+                outputString = hungarianString;
+                break;
+            //This case is to convert all Irish words
+            case "ga":
+            case "ga-IE":
+                char c = wordSample.charAt(1);
+                if (wordSample.startsWith("n") || wordSample.startsWith("t")) {
+                    if (c == 'A' || c == 'E' || c == 'I' || c == 'O' || c == 'U' || c == 'Á' || c == 'É' || c == 'Í' || c == 'Ó' || c == 'Ú') {
```
rebelDi, line 71: Can we use the replace() function here or just combine these replacements to the array and then check if the c can be replaced?

We can do it like this
char[] replacementLettersIrish = {'', '', ''}
for (int i = 0; i < letters.length; i++){
               if (c == letters[i]){
                    String IrishString1 = wordLowerCase.charAt(0) + "-" + wordLowerCase.substring(1);
               }
}

S21/rkarmuri/ConvertToLowerCase.java
```diff
@@ -0,0 +1,98 @@
...
+                break;
+            //This case is to convert Hungarian words
+            case "hu":
+                System.out.println("String after converting to lower case: " + hungarianString);
+                outputString = hungarianString;
+                break;
+            //This case is to convert all Irish words
+            case "ga":
+            case "ga-IE":
+                char c = wordSample.charAt(1);
```
rebelDi, line 69: Can we change the name of the variable or it is not neccessary?

S21/rkarmuri/Main.java
```diff
@@ -0,0 +1,21 @@
...
+public class Main {
+
+    public static void main(String[] args) {
+
+        Scanner myWord = new Scanner(System.in);//Input word
+        String wordSample;
+        Scanner myLang = new Scanner(System.in);//Input language tag
+        String language;
+        System.out.println("Enter a word to convert to lower case:");
+        wordSample = myWord.nextLine();//Move to next line after entering the word
```
rebelDi, line 15: Can we not divide the initialization and assignment of the variable? We can do this and it will make it more readable:
String wordSample = myWord.nextLine();

S21/rkarmuri/Main.java
```diff
@@ -0,0 +1,21 @@
...
+        System.out.println("Enter a language tag:");
+        language = myLang.nextLine();//Move to next line after entering the language
+        ConvertToLowerCase convertToLowerCase = new ConvertToLowerCase();
+        convertToLowerCase.convert(wordSample, language);//Calling the method to execute
```
rebelDi, line 19: Since we do not need the convertToLowerCase object anywhere else, can we just do this?
new ConvertToLowerCase().convert(wordSample, language);

rebelDi (approved):
The code works, tests are successful. Overall good code with a few minor remarks. Good job!

S21/rkarmuri/Main.java
rkarmuri: Yes we can do that.

S21/rkarmuri/ConvertToLowerCase.java
rkarmuri: Yes you can.

S21/rkarmuri/ConvertToLowerCase.java
rkarmuri: That is a good approach.

S21/rkarmuri/ConvertToLowerCase.java
rkarmuri: No we cannot have a dot at the beginning.

S21/rkarmuri/ConvertToLowerCase.java
rkarmuri: It is a good approach to put all the letters in an array, but to read and run the arrays list makes it hard for me.

S21/rkarmuri/ConvertToLowerCase.java
rkarmuri: We can do that, but I wanted these variables for each of these exceptional languages to be used write the conditions.

PR #27, opened 2021-03-11
Program and test code for converting a word to lower case

pnklshh:
Added program as well as test code.
Program can be run for checking single test case or all test cases at once.
Change the value of variable runAllTests accordingly.

### DIFF
--- a/S21/pnklshh/TestAssignment/Program.cs
+++ b/S21/pnklshh/TestAssignment/Program.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace TestAssignment
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.OutputEncoding = System.Text.Encoding.UTF8;
+            var runAllTests = false;    // To run all tests together, change the value to true
+
+            if (runAllTests)
+                Test.PerformTest();
+            else
+            {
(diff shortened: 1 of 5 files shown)


S21/pnklshh/TestAssignment/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+        public static string ConvertToLowercase(string word, string language)
+        {
+            var charList = new List<char>(word);
+            var lowerCharList = new List<char>();
+            var lowerWord = "";
+            var culture = new CultureInfo(language);
+            var vowelList = new List<char>() { 'A', 'E', 'I', 'O', 'U', 'Á', 'É', 'Í', 'Ó', 'Ú' };
```
cihanozluk, line 33: great idea instead of using all letters

S21/pnklshh/TestAssignment/Test.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace TestAssignment
+{
+    class Test
+    {
+        public static void PerformTest()
+        {
+            Console.OutputEncoding = System.Text.Encoding.UTF8;
+
+            Console.WriteLine("Enter the path of tsv file: ");
```
cihanozluk, line 14: we should get test path directly . it may be hard for the user writing path

cihanozluk (commented):
it seems nice

PR #29, opened 2021-03-12
Initial Commit

osamaibraheem11:
Change List:
1. Added a function to take in a word and language
2. Created a for loop to read from the tsv file for automated testing
3. Currently function isn't working for the Irish language (Sorry to disappoint Dr. Scannell, I know that's your favorite!)

### DIFF
--- a/S21/osamaibraheem11/s21.py
+++ b/S21/osamaibraheem11/s21.py
@@ -1,22 +1,22 @@
 import csv
 
-vowels = ["A","E","I","O","U","Á","É","Í","Ó","Ú"]
+irishVowels = ["A","E","I","O","U","Á","É","Í","Ó","Ú"]
 
 def LowerCase(word,language): 
-  if ('ga' in language) and (word.startswith('n') or word.startswith('t')) and word[1:2] in vowels:
-    return word[:1].lower()+"-"+word[1:].lower()
-  if 'en' in language:
-    return word.lower()
-  if ('tr' or "az" in language):
-    return word.replace('I',"ı").lower()
-  if ('el' in language):
-    if word.endswith('Σ'):
-      return word.replace('Σ',"ς").lower()
+    if ('ga' == language or 'ga-IE' == language) and (word.startswith('n') or word.startswith('t')) and word[1:2] in irishVowels:
+        return word[:1].lower()+"-"+word[1:].lower()
+    elif ('tr' == language):
+        return word.replace('I',"ı").lower()
+    elif ("az" == language):
+        return word.replace('I',"ı").lower()
+    elif ('el' == language):
+        if word.endswith('Σ'):
+            return word[:-1].lower() + "ς"
+        else:
+            return word[:-1].lower() + "σ"
     else:
-      return word.replace('Σ',"σ").lower()
-  else:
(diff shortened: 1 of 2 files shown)


S21/osamaibraheem11/s21.py
```diff
@@ -0,0 +1,26 @@
+import csv
+
+vowels = ["A","E","I","O","U","Á","É","Í","Ó","Ú"]
+
+def LowerCase(word,language): 
+  if ('ga' in language) and (word.startswith('n') or word.startswith('t')) and word[1:2] in vowels:
```
kscanne, line 6: Testing with "in" here will pick up languages other than Irish, since the "ga" could appear anywhere in the string. For example "nga" is the language code for Ngbaka.

S21/osamaibraheem11/s21.py
```diff
@@ -0,0 +1,26 @@
+import csv
+
+vowels = ["A","E","I","O","U","Á","É","Í","Ó","Ú"]
```
kscanne, line 3: Could we rename this as "irishVowels"?

S21/osamaibraheem11/s21.py
```diff
@@ -0,0 +1,26 @@
...
+
+def LowerCase(word,language): 
+  if ('ga' in language) and (word.startswith('n') or word.startswith('t')) and word[1:2] in vowels:
+    return word[:1].lower()+"-"+word[1:].lower()
+  if 'en' in language:
```
kscanne, line 8: Why the special case here if the default else case does word.lower()?

S21/osamaibraheem11/s21.py
```diff
@@ -0,0 +1,26 @@
...
+    return word.lower()
+  if ('tr' or "az" in language):
```
kscanne, line 10: Should separate into two boolean expressions ('tr' in language or 'az' in language).   (Noting of course that the "in" needs to be fixed as well per the comment above).

S21/osamaibraheem11/s21.py
```diff
@@ -0,0 +1,26 @@
...
+    return word.replace('I',"ı").lower()
+  if ('el' in language):
+    if word.endswith('Σ'):
```
kscanne, line 13: I'm worried about a word that ends with Σ, but also has other Σ's earlier in the word.  Since replace is global, won't those also be lowercased to ς?

S21/osamaibraheem11/s21.py
```diff
@@ -0,0 +1,26 @@
...
+      return word.replace('Σ',"ς").lower()
+    else:
+      return word.replace('Σ',"σ").lower()
+  else:
+    word.lower()
+read_tsv = csv.reader(open("tests.tsv"), delimiter="\t")
```
kscanne, line 19: Thanks for providing test code!  I'd recommend expanding the test cases, especially with tests that will catch the issues I raised above, and then re-running once those are fixed.

S21/osamaibraheem11/s21.py
Rongxiang-Zhang: Yeah, maybe "irishVowels" is more clear.

PR #30, opened 2021-03-12
intial commit with functionality

indraneelL:


S21/indraneelL/wordConvert.py
```diff
@@ -0,0 +1,29 @@
...
+            self.methodToProcessText(word,lang)
+        textFile.close()
+
+    def makeWordsLower(self,word):
+        print(word.lower())
+
+    def methodToProcessText(self, text, lang):
+        # using regex check for the language and do functionalities
+        word = text
+        if re.findall("\A[n,t][AEIOUÁÉÍÓÚ]",text) and re.match("^ga.*",lang):
```
ajayrazz99, line 22: Good to see you use an array to check the vowels all at once instead of assigning them to variables. Very efficient

S21/indraneelL/wordConvert.py
```diff
@@ -0,0 +1,29 @@
...
+        textFile = open("tests.tsv", "r")
+        while textFile:
+            line = textFile.readline()
+            if line == "" or line == '\n':
+                break
+            word, lang, _  = line.split('\t')
+            self.methodToProcessText(word,lang)
+        textFile.close()
+
+    def makeWordsLower(self,word):
```
ajayrazz99, line 16: Good use of camel casing. Makes the code readable.

S21/indraneelL/wordConvert.py
```diff
@@ -0,0 +1,29 @@
+import locale
+import re
```
ajayrazz99, line 2: Importing Regular expression certainly makes life easier

S21/indraneelL/wordConvert.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def makeWordsLower(self,word):
+        print(word.lower())
+
+    def methodToProcessText(self, text, lang):
+        # using regex check for the language and do functionalities
+        word = text
+        if re.findall("\A[n,t][AEIOUÁÉÍÓÚ]",text) and re.match("^ga.*",lang):
+            word = re.sub(r'(^[n,t])([AEIOUÁÉÍÓÚ])', r'\1-\2',text)
+        elif re.match("^tr.*", lang):
```
ajayrazz99, line 24: Although it doesn't change the functionality of the code. Traditional Python writing suggests us to use elif between if and else.  And could you please add a comment explaining the this line of code?

S21/indraneelL/wordConvert.py
```diff
@@ -0,0 +1,29 @@
...
+        textFile.close()
+
+    def makeWordsLower(self,word):
+        print(word.lower())
+
+    def methodToProcessText(self, text, lang):
+        # using regex check for the language and do functionalities
+        word = text
+        if re.findall("\A[n,t][AEIOUÁÉÍÓÚ]",text) and re.match("^ga.*",lang):
+            word = re.sub(r'(^[n,t])([AEIOUÁÉÍÓÚ])', r'\1-\2',text)
```
ajayrazz99, line 23: Efficient use of regular expressions here. Dividing the the word into two subparts and inserting a hyphen there.

S21/indraneelL/wordConvert.py
```diff
@@ -0,0 +1,29 @@
...
+        elif re.match("^tr.*", lang):
+            word = text.replace("I","ı")
+        self.makeWordsLower(word.lower())
+
+wordConvert = WordConvert()
+wordConvert.readTestFile()
```
ajayrazz99, line 29: I believe there's no need to need to write  a code for the exception of "greek" language. Could you please add a comment explain why.

S21/indraneelL/wordConvert.py
indraneelL: Thanks for the comment. Really using regular expressions was handy !!

S21/indraneelL/wordConvert.py
indraneelL: Yes, here we have used regex to find the first small letter and following with any of the vowel in the regex [] operator and making sure to taking the first letter and second letter into the subset and adding a hyphen in between them.

S21/indraneelL/wordConvert.py
indraneelL: Thanks, Whilst writing this code i have discovered more and more about Regex.

S21/indraneelL/wordConvert.py
indraneelL: Yes, Since Python 3.0, all strings are stored as Unicode in an instance and adding functionality for Σ and σ does the same as default, so i am getting the desired output of the column 3 and so, i have removed the functionality for greek and other languages.

S21/indraneelL/wordConvert.py
```diff
@@ -0,0 +1,29 @@
...
+
+## read the file and iterate through every line
+class WordConvert:
+    def readTestFile(self):
+        textFile = open("tests.tsv", "r")
+        while textFile:
+            line = textFile.readline()
+            if line == "" or line == '\n':
+                break
+            word, lang, _  = line.split('\t')
```
ajayrazz99, line 12: maybe add the 3rd column column instead of using a filler? It makes code readable.

ajayrazz99, line 12: maybe add the 3rd column column instead of using a filler? It makes code readable.

ajayrazz99 (changes requested):

PR #32, opened 2021-03-12
Needs approval for merging into the main repository

harisalam:


S21/harisalam/lowerCase.py
```diff
@@ -0,0 +1,46 @@
...
+        if res != processed_result:
+            print("""Test failed for word '{}' the result is '{}' where the expected result was '{}' """.format(
+                word, processed_result, res))
+            test_passed = False
+        else:
+            print("""Test Passed for word '{}' the result is '{}' where the expected result was '{}'""".format(
+                word, processed_result, res))
+            test_passed = True
+    if test_passed:
+        print('congratutlations all tests have')
```
myadlapalli: Even if one test case fails it is returning true

S21/harisalam/lowerCase.py
```diff
@@ -0,0 +1,46 @@
...
+
+    if (re.search('^ga.*', lan)):
+        vowels = 'A,E,I,O,U,Á,É,Í,Ó,Ú'.split(',')
+        if ((re.search('^[n,t].*', word)) and (word[1] in vowels)):
+            return((word[0]+'-'+word[1:]).lower())
+        else:
+            return(word.lower())
+
+    if (re.search('^el.*', lan)):
+        if(re.search('Σ$', word)):
```
myadlapalli, line 26: Test failed for word 'ΠΌΛΣΗΣ' the result is 'πόλςης' where the expected result was 'πόλης'Test failed for word 'ΠΌΛΣΗΣ' 
This will not work if sigma is present in the end as well as the somewhere in the middle

S21/harisalam/lowerCase.py
```diff
@@ -0,0 +1,58 @@
...
+            replace = word.replace('Σ', u"\u03C2")
+            return(replace.lower())
+        else:
+            return(word.lower())
+    if (re.search('^zh.*', lan) or re.search('^ja.*', lan) or re.search("^th.*", lan)):
```
myadlapalli, line 31: Maybe this could be the first return statement instead of all the processing

PR #33, opened 2021-03-12
Code for word record multilingual

ajayrazz99:


S21/ajayrazz99/wordChanger.py
```diff
@@ -0,0 +1,48 @@
+def LowerCaseConverter(word,lang):
```
indraneelL, line 1: It would be better if a comment is added above the method to know what this functionality is for.

S21/ajayrazz99/wordChanger.py
```diff
@@ -0,0 +1,48 @@
...
+    lang=lang
+    word=word
+    if(lang=='turkey'or lang=='az' or lang=='tr'):      #converts uppercase to lowercase
```
indraneelL, line 4: This can be optimised to if(lang=='turkey' or 'az' or 'tr'):

S21/ajayrazz99/wordChanger.py
```diff
@@ -0,0 +1,48 @@
...
+        for letter in word:
+            if(letter=='i'or letter=="I"):
+                I=u"\u0131"
```
indraneelL, line 7: what is u"\u0131" ? If a comment is added for explanation of, then it would be great.

S21/ajayrazz99/wordChanger.py
```diff
@@ -0,0 +1,48 @@
...
+                print(I,end='')
+            else:
+                print(letter.lower(),end='')
+    
+    if(lang=='Greek' or lang=='el'):
+        leng=len(word)-1
```
indraneelL, line 13: This is traditional way to iterate through the word, but python can iterate through letters of the word also. If it is needed you can optimise the code just by iterating through the words.

S21/ajayrazz99/wordChanger.py
```diff
@@ -0,0 +1,48 @@
...
+                print(I,end='')
+            else:
+                print(word[i].lower(),end='')
+        if(word[-1]=="Σ"):  
+            I=u"\u03C2"
+            print(I,end='')
+    
+    if(lang=='irish' or lang=='ga' or lang=='ga-IE' or lang=='Irish'):
+        leng=len(word)
+
```
indraneelL, line 26: May be we can consider all the vowels into an array and as we already have the word, for looping vowels array and adding '-' at desired location would be great. what do you think ?

S21/ajayrazz99/wordChanger.py
```diff
@@ -0,0 +1,48 @@
...
+    
+        if (word[0]=='n' or word[0]=='t'):
+            if(word[1]=='A' or word[1]=='E' or word[1]=='I' or
+            word[1]=='O' or word[1]=='U' or word[1]=='Á' or word[1]=='É' or word[1]=='Í' or word[1]=='Ó'
+            or word[1]=='Ú'):
+                print(word[0].lower()+'-'+word[1].lower(),end='')
+                print(word[2:].lower())        
+        else:
+            print(word.lower())
+
```
indraneelL, line 35: all the below if conditions have same functionality 
```suggestion
            for letter in word:
             letter=word.lower()
         print(letter)
```
may be you can add them to a only one if condition. Will that would be great and minimize the code ?

indraneelL (changes requested):

S21/ajayrazz99/wordChanger.py
ajayrazz99: yes, I agree. And the function's name was pretty much self explanatory but as you suggested adding a comment will surely help understand. Thanks!

S21/ajayrazz99/wordChanger.py
ajayrazz99: Yes, thank you. I'll make sure to change that.

S21/ajayrazz99/wordChanger.py
ajayrazz99: That certain piece of expression is the universal code or uni-code for the symbol 'Σ' but surely a comment would be great. Thanks!

S21/ajayrazz99/wordChanger.py
ajayrazz99: Yes, good suggestion. I'll apply the changes. Thank you!

S21/ajayrazz99/wordChanger.py
ajayrazz99: That is exactly what I did. But the difference is that each and every vowel is assigned to a variable.

PR #35, opened 2021-03-12
Word lowercase test

Rongxiang-Zhang:


S21/Rongxiang-Zhang/lower.py
```diff
@@ -0,0 +1,60 @@
+import csv
+
+
+def wordlower(word,lan):
+
+    word=word
+
+    lan=lan
```
osamaibraheem11, line 8: We don't need to reassign the variables to the same name, we can just use the names we passed into the function. For example, 
def NamePrinter (name):
     print(f'Hello! {name}')

S21/Rongxiang-Zhang/lower.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if(lan=='en'):
+        for letter in word:
```
osamaibraheem11, line 11: We don't need to use a for loop here, this will add complexity. You can use return or print word.lower(). It will take care of all the word as a whole

S21/Rongxiang-Zhang/lower.py
```diff
@@ -0,0 +1,60 @@
...
+
+        for letter in word:
+            if (letter=='i' or letter=='I'):
+                letter=word.replace('I',"ı").lower()
+                print(letter)
+            else:
+                print(letter.lower(),end='')
+
+
+    if(lan=='ga'):
```
osamaibraheem11, line 26: How do we account for ga-IE? Current this if statement will not count ga-IE into this logic. We could say if "ga" in lan, this make it take into account all "ga" that have the same logic

S21/Rongxiang-Zhang/lower.py
```diff
@@ -0,0 +1,60 @@
...
+        leng=len(word)
+
+        if(word[0]=='n' or word[0]=='t'):
+            if(word[1]==a or word[1]==E or word[1]==I or word[1]==O or word[1]==U or word[1]==Á or word[1]==Í or word[1]==Ó or word[1]==Ú):
```
osamaibraheem11, line 30: This if statement is long, we could put all the vowels in a list. We can say if word[1] in list of vowels, run this logic

S21/Rongxiang-Zhang/lower.py
```diff
@@ -0,0 +1,60 @@
...
+
+                for letter in word:
+                    member = ['-']
+                    word.insert(1,member)
+                    print(word.lower(),end='')
+            else:
+                print(word.lower()，end='')
+
+    if(lan=='el'):
+        leng=len(word)
```
osamaibraheem11, line 40: We have this variable named leng in two of our if statements, but they never get used. We can get ride of them, if they aren't used.

S21/Rongxiang-Zhang/lower.py
```diff
@@ -0,0 +1,60 @@
+import csv
+
+
+def wordlower(word,lan):
```
osamaibraheem11, line 4: Good names! for the function and word!

S21/Rongxiang-Zhang/lower.py
```diff
@@ -0,0 +1,60 @@
...
+                print(letter)        
+
+        else:
+            print(word.lower(), end='')
+
+    if(lan=='th' or lan=='zh'):
+        for letter in word:
+            letter=word.lower()
+        print(letter)
+
```
osamaibraheem11, line 55: There is a lot of empty space that can sometimes be helpful we are reading and writing code but I think we can get down on some of the spacing. 

Also, currently we just a function and it is not being used. It would be great if we could have a for loop to go through the test.tsv file and test out each input! This could help us have faster code reviews and testing!

osamaibraheem11 (changes requested):
See comments below.

S21/Rongxiang-Zhang/lower.py
Rongxiang-Zhang: Yes, you're right.

S21/Rongxiang-Zhang/lower.py
Rongxiang-Zhang: well,I have to admit I made it looks more complicated.

S21/Rongxiang-Zhang/lower.py
Rongxiang-Zhang: I wanted to give a variable to every vowel, but it seemed too long.

S21/Rongxiang-Zhang/lower.py
Rongxiang-Zhang: Great idea, I will correct it.

PR #38, opened 2021-03-12
_class_Code_Review

hvkone:
ui_txt_conv.py is final


FaizanSyed95:
I see that you have uploaded the outdated version of your codes, it's a great practice to do when starting and learning about the project. However, It would be better if next time we can only upload the final version of the file so that it might be easier for the reviewers to understand which file they have to review.

S21/hvkone/ui_txt_conv.py
```diff
@@ -0,0 +1,81 @@
+#---------------------------------
+# DEPENDECIES
+#---------------------------------
+# pip install langdetect
+from langdetect import detect
+
+# pip install translate
+from translate import Translator
+
+# pip install pycountr
```
FaizanSyed95, line 10: It nice that you have explained about the libraries you installed, however; for this code, we think this is something more and time-consuming, it can be done even without installing these libraries. 
Also, it was discussed not to use built-in libraries for this project.

S21/hvkone/ui_txt_conv.py
```diff
@@ -0,0 +1,81 @@
...
+import pycountry
+
+# pip install bcp47
+from bcp47 import bcp47
+
+# pip install langcodes
+# pip install language_data
+from langcodes import *
```
FaizanSyed95, line 18: This again needs a library to install first to run the code. For us, it does not look like a good idea to install so many libraries because the code reviewer has to install all these libraries to run and test the code. Is it possible if we can change the code and run it without these libraries?

S21/hvkone/ui_txt_conv.py
```diff
@@ -0,0 +1,81 @@
...
+    if(usr_word[i] >= 'A' and usr_word[i] <= 'Z'):
+        usr_asc_new = usr_asc_new + chr((ord(usr_word[i]) + 32))
+    else:
+        usr_asc_new = usr_asc_new + usr_word[i]
+usr_asc_new = convert_to_ascii(usr_asc_new)       
+print('CONVERTED ASCII OUTPUT: ', usr_asc_new)
+
+usr_lang = input('Type Language Code [en]: ')
+usr_regn = input('Type Region Code [US]: ')
+
```
FaizanSyed95, line 45: It would be beneficial if we read files, instead of giving inputs individually. Because there might be multiple test cases, and it would look time-consuming if we put them individually in the code.

S21/hvkone/ui_txt_conv.py
```diff
@@ -0,0 +1,81 @@
...
+#---------------------------------
+# LANGUAGE CONDITIONALS
+#---------------------------------
+
+# IF GAELIC INPUT
+if lang_code == 'ga' and (usr_word.startswith('t') or usr_word.startswith('n')) and usr_word[1] in ['A','E','I','O','U','Á','É','Í','Ó','Ú']:
+    conv_word = '-'.join([usr_word[0],usr_word[1:]])
+    print('Hyphenated Lowercase Conversion: ', conv_word.lower())
+else:
+    conv_word = usr_word_code.lower()
```
FaizanSyed95, line 71: Does this else condition include other languages such as Chinese, Thai, and English? 
Because only English can be converted to lower case, Chinese , Thai looks the same.

S21/hvkone/ui_txt_conv.py
```diff
@@ -0,0 +1,81 @@
+#---------------------------------
```
FaizanSyed95, line 1: Overall it looks fine, your ideas are great. I would suggest adding more comments on your individual functions to help the reviewers understand them more.

S21/hvkone/ui_txt_conv.py
hvkone: Okay, thank you. I was just trying to see if I could input various languages for detection.

S21/hvkone/ui_txt_conv.py
hvkone: No... I was using that for each condition. I should probably use that line once, for all of the language conditions. Thank you, I didn't think of that.

S21/hvkone/ui_txt_conv.py
hvkone: Will do! Thank you for the review.

S21/hvkone/ui_txt_conv.py
FaizanSyed95: No problem, that's great we understand each other points.

S21/hvkone/ui_txt_conv.py
FaizanSyed95: Yes, that would be great.

S21/hvkone/ui_txt_conv.py
FaizanSyed95: Your welcome and thank you for understanding my points.

PR #59, opened 2021-04-07
Refactored code after team discussion

indraneelL:
Team performed refactoring :
Indraneel, Raju Karmuri, Garrett Reed, Sai Sri Raj.

### DIFF
--- a/S21/indraneelL/refactorme.py
+++ b/S21/indraneelL/refactorme.py
@@ -1,84 +1,86 @@
 import unicodedata
 
+class BaseClass:
+    CAPITAL_ACUTE_A = '\u00c1'
+    CAPITAL_ACUTE_E = '\u00c9'
+    CAPITAL_ACUTE_I = '\u00cd'
+    CAPITAL_ACUTE_O = '\u00d3'
+    CAPITAL_ACUTE_U = '\u00da'
+    CAPITAL_I = '\u0049'
+    DOTLESS_i = '\u0131'
+    CAPITAL_SIGMA = '\u03a3'
+    SMALL_SIGMA = '\u03c2'
+
+    STANDARD_VOWELS = 'AEIOU' + CAPITAL_ACUTE_A + CAPITAL_ACUTE_E + CAPITAL_ACUTE_I + CAPITAL_ACUTE_O + CAPITAL_ACUTE_U
+
+    def __init__(self,word):
+        self.word = word
+
+class CheckIrishWord(BaseClass):
+
+    def process_word(self):
+        temp = self.word
+        if len(self.word)>1:
+            if (self.word[0]=='t' or self.word[0]=='n') and unicodedata.normalize('NFC', self.word)[1] in self.STANDARD_VOWELS:
+                temp = self.word[0]+'-'+temp[1:]
+        return temp.lower()
+
+class CheckTurkAndAzerWord(BaseClass):
+
+    def process_word(self):
+        temp = self.word
+        temp = temp.replace(self.CAPITAL_I,self.DOTLESS_i)
+        return temp.lower()
+
+class CheckGreekWord(BaseClass):
+
+    def process_word(self):
+        temp = self.word
+        if temp[-1]== self.CAPITAL_SIGMA:
+            temp = temp[:-1]+ self.CAPITAL_SIGMA
+        return temp.lower()
+
 class Word:
 
-  def __init__(self, word, bcpCode, std=False):
-    self._w = word
-    self._l = bcpCode
-    self._finalSigma = False
-    self._standardIrishSpelling = std
-    # OLD EXPERIMENTAL CODE for dealing with vowel harmony
-    # self._numVowels = 0
-    # for c in word:
-    #   if c in 'aeiouAEIOU':
-    #   self._numVowels += 1
+    def __init__(self, word, lang):
+        self._w = word
+        self._l = lang
 
-  def setWord(self, w):
-    self._w = w
+    def checkForLangCode(self):
+        language = self._l
+        if '-' in self._l:
+            i = self._l.find('-')
+            language = self._l[0:i]
+        if language == "az" or language == "tr":
+                language = "at"
+        return language
 
-  def toLower(self):
-    language = self._l
-    if '-' in self._l:
-      i = self._l.find('-')
-      language = self._l[0:i]
-    if len(language)<2 or len(language)>3:
-      print("Invalid BCP-47 code")
-      return ''
-    temp = self._w
-    if language=='zh':
-      return temp
-    elif language=='ja':
-      return temp
-    elif language=='ga':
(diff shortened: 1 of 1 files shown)


indraneelL:
yes, I will replace the temp variable, Even Raju also suggested this. Thanks

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,86 @@
...
+
+    STANDARD_VOWELS = 'AEIOU' + CAPITAL_ACUTE_A + CAPITAL_ACUTE_E + CAPITAL_ACUTE_I + CAPITAL_ACUTE_O + CAPITAL_ACUTE_U
+
+    def __init__(self,word):
+        self.word = word
+
+class CheckIrishWord(BaseClass):
+
+    def process_word(self):
+        temp = self.word
```
rkarmuri, line 22: We can replace all instances of "temp" with "word".

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,86 @@
...
   for line in f:
     line = line.rstrip('\n')
     pieces = line.split('\t')
     w = Word(pieces[0], pieces[1])
     answer = w.toLower()
-    if answer != pieces[2]:
-      print('Test case failed. Expected', pieces[2], 'when lowercasing',pieces[0],'in language',pieces[1],'but got',answer)
-  f.close()
+    if w.checkThePieces(answer, pieces):
+      raise Exception('Test case failed. Expected', pieces[2], 'when lowercasing',pieces[0],'in language',pieces[1],'but got',answer)
```
rkarmuri, line 85: Good use of exception instead of print.

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,86 @@
...
+    def __init__(self, word, lang):
+        self._w = word
+        self._l = lang
+
+    def checkForLangCode(self):
+        language = self._l
+        if '-' in self._l:
+            i = self._l.find('-')
+            language = self._l[0:i]
+        if language == "az" or language == "tr":
```
rkarmuri, line 54: I believe we need to have language tags be "ga" and "gd" instead of "az" "tr" here.

rkarmuri (changes requested):
Overall the code is running good with the refactors when I tested, I suggest some minor changes.

S21/indraneelL/refactorme.py
indraneelL: No, as the functionality for "Turkish" and "Azer" is the same I just replaced the language variable with a common one which is "at", which performs the same operation for both languages. Removed some of the languages as they are not in tests.tsv. I'm not sure if we need to refactor based on tests.tsv or consider all the languages which are already in the professor's code.

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,92 @@
 import unicodedata
 
+class BaseClass:
+    CAPITAL_ACUTE_A = '\u00c1'
```
kscanne, line 4: As I mentioned in class, it would be nice to name these according the Unicode data files, e.g. here "LATIN_CAPITAL_LETTER_A_WITH_ACUTE".  Also this is a magic constant; you can embed the literal unicode character in the Python source.

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,92 @@
...
+    CAPITAL_ACUTE_E = '\u00c9'
+    CAPITAL_ACUTE_I = '\u00cd'
+    CAPITAL_ACUTE_O = '\u00d3'
+    CAPITAL_ACUTE_U = '\u00da'
+    CAPITAL_I = '\u0049'
+    DOTLESS_i = '\u0131'
+    CAPITAL_SIGMA = '\u03a3'
+    SMALL_SIGMA = '\u03c2'
+
+    STANDARD_VOWELS = 'AEIOU' + CAPITAL_ACUTE_A + CAPITAL_ACUTE_E + CAPITAL_ACUTE_I + CAPITAL_ACUTE_O + CAPITAL_ACUTE_U
```
kscanne, line 14: Do all these special cases need to be in the base class?

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,92 @@
...
+
+class CheckGreekWord(BaseClass):
+
+    def process_word(self):
+        _word = self.word
+        if _word[-1]== self.CAPITAL_SIGMA:
+            _word = _word[:-1]+ self.CAPITAL_SIGMA
+        return _word.lower()
+
 class Word:
```
kscanne, line 49: Didn't expect to see this since you already have BaseClass.  What's the connection between Word and BaseClass?  Combine them?

kscanne: oh I think I understand better now.   You might consider renaming BaseClass to something like "LowercaseStrategy" since you're lumping together the algorithms, independent of language, in its subclasses.  (Also move the aliases for special characters into subclasses).

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,92 @@
...
-      else:
-        return self._w[0].lower() in 'bcdfgmpst' and self._w[1].lower()=='h'
-    else:
-      raise NotImplementedError('Method only available for Irish and Scottish Gaelic')
+    def __init__(self, word, lang):
+        self._w = word
+        self._l = lang
+
+    def checkForLangCode(self):
+        language = self._l
```
kscanne, line 56: Might prefer this functionality in a separate class representing BCP-47 strings.

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,92 @@
...
+        if '-' in self._l:
+            i = self._l.find('-')
+            language = self._l[0:i]
+        if language == "az" or language == "tr":
+                language = "at"
```
kscanne, line 61: This is a hack and not worth it.  Just include separate entries in the switcher.  It happens that there's no ISO 639-1 code "at", but that was just luck!

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,92 @@
...
+        return language
+
+    def switcherProcess(self,language):
+        
+        switcher = {
+            "ga": CheckIrishWord(self._w).process_word(),
+            "at": CheckTurkAndAzerWord(self._w).process_word(),
```
kscanne, line 68: See comment above.

S21/indraneelL/refactorme.py
```diff
@@ -1,84 +1,92 @@
...
+            "el": CheckGreekWord(self._w).process_word(),
+        }
+
+        return switcher.get(language, self._w.lower())
```
kscanne, line 72: This doesn't handle zh/ja/th correctly since you're calling lower() in those cases and we don't want to.

PR #61, opened 2021-04-08
Refactored refactorme.py

hvkone:


S21/hvkone/refactorme.py
```diff
@@ -2,83 +2,57 @@
...
-    # OLD EXPERIMENTAL CODE for dealing with vowel harmony
-    # self._numVowels = 0
-    # for c in word:
-    #   if c in 'aeiouAEIOU':
-    #   self._numVowels += 1
+  def __init__(self, word, language):
+    self.word = word 
+    self.lang = language
+  
+  def get_language(self):
```
kscanne, line 9: Might be nice to separate out this functionality into a class representing BCP-47 strings

S21/hvkone/refactorme.py
```diff
@@ -2,83 +2,57 @@
...
+    language = self.lang
+    
+    if '-' in self.lang:
+      index = self.lang.find('-')
+      language = self.lang[0:index]
+      return language
+      
+    if not len(language)==2:
```
kscanne, line 17: These can have language codes of length 3 also (See ISO 639-3)

S21/hvkone/refactorme.py
```diff
@@ -2,83 +2,57 @@
...
+      raise ValueError("Invalid BCP-47 code")
+    return language
+    
+  def change_to_lower(self):
+    language=self.get_language()
+    not_lower_list = ('zh','ja','th')
```
kscanne, line 23: Still not scalable to dozens or hundreds of "non-casing" languages.

S21/hvkone/refactorme.py
```diff
@@ -2,83 +2,57 @@
...
+    Irish_vowels = ('AEIOU\u00c1\u00c9\u00cd\u00d3\u00da')
```
kscanne, line 24: Magic constants?  You can embed literal unicode characters in Python source.

S21/hvkone/refactorme.py
```diff
@@ -2,83 +2,57 @@
...
+    
+    if language in not_lower_list:
+      return self.word
 
-  def setWord(self, w):
-    self._w = w
+    elif language==('ga' or 'gd'):
```
kscanne, line 29: Much cleaner, but is there a design that avoids this cascade of if/then/else on every call to lowercase?

kscanne, line 29: syntax is wrong here; this is checking if language equals-equals the boolean on the RHS.  Did you run tests?

S21/hvkone/refactorme.py
```diff
@@ -2,83 +2,57 @@
...
+      if len(self.word)>1 and (self.word[0]=='t' or self.word[0]=='n') and unicodedata.normalize('NFC', self.word)[1] in Irish_vowels:
+        self.word = self.word[0]+'-'+self.word[1:]
+      return self.word.lower()
+
+    elif language==('tr' or 'az'):
+      self.word = self.word
```
kscanne, line 35: ?

S21/hvkone/refactorme.py
```diff
@@ -2,83 +2,57 @@
...
+    Irish_vowels = ('AEIOU\u00c1\u00c9\u00cd\u00d3\u00da')
+    
+    if language in not_lower_list:
+      return self.word
 
-  def setWord(self, w):
-    self._w = w
+    elif language==('ga' or 'gd'):
+      if len(self.word)>1 and (self.word[0]=='t' or self.word[0]=='n') and unicodedata.normalize('NFC', self.word)[1] in Irish_vowels:
+        self.word = self.word[0]+'-'+self.word[1:]
```
kscanne, line 31: this method should not change the string itself, should just return the lowercase version.

PR #63, opened 2021-10-05
Adding the files for code review

cemilcan98:
Added the files and test cases for the code review


cemilcan98:
Upload test cases for the triangle problem

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
+def toLowerCase(i):
```
kscanne, line 1: I would suggest a more descriptive parameter name here.

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
...
+            y = y[:-1] + '\u03C2'
+        y = y.lower()
+        return y
+
+    if(i[1] in ["zh", "ja", "th", "zh-Hans"]):
+        y = i[0]
+        return y
+
+
+def open_file(file_name):
```
kscanne, line 64: Perhaps "read_file" or "parse_file"?  I was confused below when this was called, expecting "y" to be a file object.

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
...
+    return pieces_text
+
+
+def close_file(file_name):
+    f = open(file_name, encoding="utf8")
+    f.close()
+
+
+y = open_file('tests.tsv')
+for k in y:
```
kscanne, line 82: Clearer variable names would help here as well.

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
...
+    pieces_text = []
+    for line in f:
+        line = line.rstrip('\n')
+        lines_text.append(line)
+        pieces = line.split('\t')
+        pieces_text.append(pieces)
+    return pieces_text
+
+
+def close_file(file_name):
```
kscanne, line 76: Not sure I understand this function.

F21/cemilcan98/app.py
kscanne: I like that you separated the reading of the file into its own function, but the rest of the code is tightly coupled to the form of the data structure that you return from that function.

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
+def toLowerCase(i):
+    if(i[1] in ["en", "en-US", "en-IE", "en-Latn"]):
```
kscanne, line 2: The code should be much more general than just the examples in the test file. Worth reading through the BCP-47 standard for this.

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
+def toLowerCase(i):
+    if(i[1] in ["en", "en-US", "en-IE", "en-Latn"]):
+        y = i[0].lower()
+        return y
+
+    # Checking for the special letters in Turkish and Azeri 
+    if(i[1] == "tr" or i[1] == "az"):
+        y = i[0]
+        y = y.replace('\u011E', '\u011F')
```
kscanne, line 9: The Unicode code points are not very readable here; either use the explicit characters or explain what these are in comments.

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
...
+        y = y.replace('\u0049', '\u0131')  # I to ı
+        y = y.lower()
+
+        return y
+
+    # Checking for the special cases in Irish
+    if(i[1] == "ga" or i[1] == "ga-IE"):
+        y = i[0]
+        first = i[0][0]
+        second = i[0][1]
```
kscanne, line 23: Crash if length of i[0]<2?

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
...
+        first_letters = ['n', 't']
+        second_letters = ['A', 'E', 'I', 'O', 'U',
+                          '\u00C1', '\u00C9', '\u00CD', '\u00D3', '\u00D3']
```
kscanne, line 26: Same comment regarding the unicode escapes.

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Greek Cases
+    if(i[1] == "el"):
+        y = i[0]
+        if(y[-1] == '\u03A3'):
+            y = y[:-1] + '\u03C2'
+        y = y.lower()
+        return y
+
+    if(i[1] in ["zh", "ja", "th", "zh-Hans"]):
```
kscanne, line 59: Is this scalable to a list of hundreds of languages where case is irrelevant?

F21/cemilcan98/app.py
```diff
@@ -0,0 +1,84 @@
+def toLowerCase(i):
+    if(i[1] in ["en", "en-US", "en-IE", "en-Latn"]):
+        y = i[0].lower()
+        return y
+
+    # Checking for the special letters in Turkish and Azeri 
+    if(i[1] == "tr" or i[1] == "az"):
```
kscanne, line 7: Would it be better to have elif's throughout since these cases are mutually exclusive?

F21/cemilcan98/app.py
cemilcan98: I have just pushed a new version of the code with considering all your comments and suggestions.

PR #70, opened 2021-10-06
initial commit - added test cases, created functions

maddiekoehler:


F21/maddiekoehler/lower.py
```diff
@@ -0,0 +1,40 @@
...
+	#Turkish or Azerbaijani special characters
+	if 'tr' in lang or 'az' in lang:
+		word = word.replace('I','ı')
+		word = word.lower()
+
+	#Irish special characters
+	irishVowels = ['A','E','I','O','U','Á','É','Í','Ó','Ú']
+	if 'ga' in lang and len(word)>1:
+		if word[0] == 'n' or word[0] == 't':
+			if word[1] in irishVowels:
```
AmirSalari, line 19: Pass   Expected: nõg   Actual: nõg
Fail   Expected: nõg   Actual: n-õg 

Both words are the same but different results. again I don't know why this happens, code seems to be correct

PR #71, opened 2021-10-06
Code review class

yzhang120:


F21/yzhang120/lower_case.py
```diff
@@ -0,0 +1,43 @@
+import pandas as pd
+
```
mfdunlap, line 2: Adding comments to your functions to give an overview of what they do would help with readability/understandability

F21/yzhang120/lower_case.py
```diff
@@ -0,0 +1,43 @@
...
+def lower_tr_az(input_string):
+    output_string = ''
+    for char in input_string:
+        if char != 'I':
+            output_string += char.lower()
+        else:
+            output_string += u"\u0131"
+    return output_string
+
+def lower_ga(input_string):
```
mfdunlap, line 18: While this code should work in theory but this code may not handle if the unicode for accented letters is decomposed unicode instead of normalized NFC unicode. This may cause your code to incorrectly handle Gaelic text if it has vowels written using two chars (" ‘+o ") instead of one char (" ó ").

F21/yzhang120/lower_case.py
```diff
@@ -0,0 +1,43 @@
...
+    Irish_vowel = 'AEIOUÁÉÍÓÚ'
+    if input_string[0] in 'nt' and input_string[1] in Irish_vowel:
+        output_string = input_string[0] + '-' + input_string[1:].lower()
+    else:
+        output_string = input_string.lower()
+    return output_string
+
+def get_lowercase(word, language):
```
mfdunlap, line 26: Since this is written using the BCP-47 specification, you can use the first two letters of the language to determine the langauge rules the lowercase function should adhere to. You may want to strip the input first to get rid of any extra white space that may be included. Then you don't have to include every single case for each language.

F21/yzhang120/lower_case.py
mfdunlap: I would look up unicode normalization.
https://docs.python.org/3/library/unicodedata.html
https://towardsdatascience.com/difference-between-nfd-nfc-nfkd-and-nfkc-explained-with-python-code-e2631f96ae6c

PR #74, opened 2021-10-06
Capital to lower case characters.

MahdiRahbar:
Here is the main code for converting capital letters to lower case. 
You can file three main ```.py``` files in the folder. ```main.py``` is responsible to run the whole program, get the data from the client and give it back to the program to give back the demanded results. ```lower.py``` is the module containing ```Lower``` class which is the core of the program and finally, ```test.py``` is responsible to get our test data and evaluate the program's output.


F21/mahdirahbar/lower.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class Lower: 
+    def __init__(self, string, lang):
+        self.string = string 
+        self.lang = lang 
+        self.detected_lang = self.lang_detector()
+        self.word_list = self.string_splitter(string)
+
+    def string_splitter(self, string):
```
abeiler7, line 17: This does exactly what you want it to do, but for future reference Python has a built in function `string.split()` that would do the same for you.

F21/mahdirahbar/lower.py
```diff
@@ -0,0 +1,155 @@
...
+            output: returns the fianl string
+        '''
+
+        if 'en' in self.lang:
+            # English 'en'
+            return self.string.lower()
+        elif 'ga' in self.lang :
+            # Irish 'ga'
+            self.ga_util()
+            return self.ga_lower()
```
abeiler7, line 148: A coding standard that I have seen in place in multiple companies is to not multiple returns in a single function. It can make it harder to trace and debug in the future. 
Rather you can assign the potential return values to a variable like `lower_string` and then return that at the end of the function.

PR #77, opened 2021-10-06
adding lowercase function, updating testcases, creating unittest file

megannigg:


F21/megannigg/lowercase.py
```diff
@@ -0,0 +1,30 @@
+def lowercase(word, language):
+
+#Turkish or Azerbaijani uses lowercase i with out a dot    
+    if 'tr' in language or 'az' in language:
+        word = word.replace('I','ı')
+        word = word.lower()
+        
+#Irish rule: when converting uppercase vowels to lowercase, the vowel must be proceded by a dash
+    elif 'ga' in language:
```
kscanne, line 9: I don't think "in" is strong enough here; for example there are three-letter codes like "dga" for the Dagaare language spoken in Ghana.

F21/megannigg/lowercase.py
```diff
@@ -0,0 +1,30 @@
...
+        word = word.lower()
+ 
+#Greek and uppercase sigma at the end of a word gets converted to a small letter final sigma
+    elif 'el' in language:
+        if word[-1] == 'Σ':
+                word = word[:-1]+'ς'
+        word = word.lower()
+
+#chinese and japanese do not have lowercase values
+    elif 'zh' in language or 'ja' in language or 'th' in language:
```
kscanne, line 22: Is this scalable to hundreds of such cases?

F21/megannigg/lowercase.py
```diff
@@ -0,0 +1,30 @@
...
+
+#Turkish or Azerbaijani uses lowercase i with out a dot    
+    if 'tr' in language or 'az' in language:
+        word = word.replace('I','ı')
+        word = word.lower()
+        
+#Irish rule: when converting uppercase vowels to lowercase, the vowel must be proceded by a dash
+    elif 'ga' in language:
+        if len(word)>1:
+            if word[0] in 'tn' and word[1] in 'AEIOUÁÉÍÓÚ':
```
kscanne, line 11: This will only work as expected on the test cases if you do some Unicode normalization first.

PR #87, opened 2021-10-06
Add files via upload

BhargavaReddy24:
BhargavaReddy  

Code review


F21/bhargavareddy24/BRCodeReview-01.py
```diff
@@ -0,0 +1,41 @@
+class WordLowerCase:
+    def init(self, word, language):
+        self.word = word
+        self.l = language
+
+    def lower(self):
```
bmacgavin, line 6: I like that you kept all the functions contained within a class that handled the replacement/lowercase in one function

PR #90, opened 2021-10-06
Creation of lowercase function

mfdunlap:
Creation of a lower case function which handles special cases for lowercasing words in various languages; Includes a series of test for functionality


F21/mfdunlap/lowercaser.py
```diff
@@ -0,0 +1,98 @@
+import csv
+import unicodedata
+
+#---------------------------------------------------------------------#
+#                         LOWERCASE FUNCTION                          #
+#---------------------------------------------------------------------#
+def lowercase(word, language):
+    """
```
yzhang120, line 8: Great documentation.

PR #92, opened 2021-10-06
Word lower case

GadeManasa:
The creation of a lowercase function that handles special cases for lowercasing words in different languages.

### DIFF
--- a/F21/gademanasa/main.py
+++ b/F21/gademanasa/main.py
@@ -1,0 +1,37 @@
+class LowerCase:
+    def __init__(self, word, bcpCode):
+        self.word = word
+        self.l = bcpCode
+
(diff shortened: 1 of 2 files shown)


F21/gademanasa/main.py
```diff
@@ -0,0 +1,37 @@
+class LowerCase:
+    def __init__(self, word, bcpCode):
+        self.word = word
+        self.l = bcpCode
+
+    def wordLower(self):
+        self.l=self.l[:2]
+        if self.l == 'tr' or self.l == 'az':
```
Dheeraj-Thikkamaneni, line 8: what do you say about comparing the strings using 'as in' keyword which is easy to read, for eg: self.l as in tr

PR #95, opened 2021-10-06
added function+output, use: py _toLowerCase.py

JoeLanzi:


JoeLanzi:
added function+output, use: py _toLowerCase.py

F21/joelanzi/Output Result.tsv
```diff
@@ -0,0 +1,19 @@
...
+nATHAIR	ga	n-athair	n-athair	True
+NATHAIR	ga	nathair	nathair	True
+nÓg	ga-IE	n-óg	n-óg	True
+nÕg	ga-IE	nõg	nõg	True
+nÕg	ga-IE	nõg	nõg	True
+KASIM	tr	kasım	kasım	True
+KASIM	en	kasim	kasim	True
+ΠΌΛΗΣ	el	πόλης	πόλης	True
+官话	zh-Hans	官话	官话	True
+ภาษาไทย	th	ภาษาไทย	ภาษาไทย	True
```
lorenzopappone, line 19: We should add more cases to this file in order to achieve wider coverage.

F21/joelanzi/_toLowerCase.py
```diff
@@ -0,0 +1,46 @@
+import pandas as pd
+from unicodedata import normalize
+
+## To Lowercase Function
+def _to_lower(word, language):
```
lorenzopappone, line 5: We should add some comments regarding the lowercase function, describing input and output parameters.

F21/joelanzi/Output Result.tsv
lorenzopappone: We should add more languages to the file, adding sentences instead of only words and punctuations.

F21/joelanzi/_toLowerCase.py
```diff
@@ -0,0 +1,46 @@
...
+    lang = language[0:language.find('-')] if '-' in language else language
+    if lang == 'tr' or lang == 'az':
+        return word.replace('I','ı').lower() # '\u0049','\u0131'
+    elif lang == 'ga':
+        irish_vowels = ['A','E','I','O','U','Á','É','Í','Ó','Ú'] # 'AEIOU\u00c1\u00c9\u00cd\u00d3\u00da'
+        norm_word = normalize('NFC',word)
```
lorenzopappone, line 11: We should describe using brief comments the normalize function.

F21/joelanzi/_toLowerCase.py
lorenzopappone, line 5: This script could be structured differently to allow better code comprehension and reuse. For instance, we could add a utility file with classes or methods used for the lowercase tasks and separate it from the main code used for running the script.

F21/joelanzi/_toLowerCase.py
```diff
@@ -0,0 +1,46 @@
+import pandas as pd
+from unicodedata import normalize
+
+## To Lowercase Function
+def _to_lower(word, language):
+    lang = language[0:language.find('-')] if '-' in language else language
+    if lang == 'tr' or lang == 'az':
```
lorenzopappone, line 7: Maybe adding a vector with all the exceptions like "tr", "az", "ga", etc., would permit an easier way to read all the exceptions instead of scrolling the script. To do this, you can add a line such as "if language is in **_vector_**, and then specify the code for each exception.

F21/joelanzi/_toLowerCase.py
```diff
@@ -0,0 +1,46 @@
...
+    df = pd.read_csv("tests.tsv", sep='\t', header=None)
+    df.columns =['word','lang','expected_lower_case']
+
+    # Cover word to lower
+    output = []
+    for (w,l) in zip(df.iloc[:, 0],df.iloc[:, 1]):
+        output.append(_to_lower(w,l))
+    df['function_output'] = output
+
+    # Returns true/false
```
lorenzopappone, line 39: true or false about what?

F21/joelanzi/myers.tsv
```diff
@@ -1,3 +1,8 @@
+{"a" : 3, "b" : 3, "c" : 3}	equilateral
+{"a" : 10, "b" : 10, "c" : 10}	equilateral
 {"a" : 3, "b" : 4, "c" : 5}	scalene
+{"a" : 5, "b" : 7, "c" : 10}	scalene
 {"a" : 3, "b" : 3, "c" : 1}	isosceles
+{"a" : 5, "b" : 5, "c" : 3}	isosceles
 {"a" : 3, "b" : 4, "c" : 100}	error
+{"a" : 2, "b" : 2, "c" : 50}	error
```
kscanne, line 8: Good tests... clearly need many more!

PR #99, opened 2021-10-06
code review

shirishma1:


F21/shirishma1/Code Review.py
```diff
@@ -0,0 +1,51 @@
...
+        return obj.Irish(text)
+    elif lang_code == 'el':
+        return obj.Greek(text) 
+    elif lang_code == 'tr':
+        return obj.Turkish(text)
+    
+    else:
+        return 'invalid language code'
+Word = input()
+Language_Code = input()
```
Pream9, line 48: For better understanding In the input ("Enter Language code:")

F21/shirishma1/Code Review.py
```diff
@@ -0,0 +1,51 @@
+class LowerCaseConversion():
+   
+    def lowercase(self,text): # if there are no specific conditions this method is used
+        return text.lower()
+        
+    def Irish(self,text):
```
Pream9, line 6: My suggestion is you have defined a function for every language it's better to define a function, It reduces the spaces and line of code .

shirishma1: Code Readability is good when a function does only one thing and if a new language comes adding another function will not break any other functions.

PR #105, opened 2021-10-06
addition of program 

Kunaixfox:


F21/annoyed-raven/main.py
```diff
@@ -0,0 +1,104 @@
...
+        language_encoding.append(transition[1])
+        lowercased.append(transition[2])       
+'''     an easy check to make sure tsv file was seperated correctly
+        print(row)
+        print(capitalized)
+        print(language_encoding)
+        print(lowercased)
+'''  
+# the storage lists of indexes for each occurence of a the specific dialact/language
+Encoding_indexesEN= []
```
GreihMurray, line 28: A 2D list or dictionary may be beneficial here as it would allow for fewer variables and could reduce the amount of for loops further down in the file when converting to lowercase. A dictionary may be more useful as it would allow for using the same variable names within the overall variable. It would also make it easier for the program to handle additional languages being added to the test file

F21/annoyed-raven/main.py
```diff
@@ -0,0 +1,104 @@
...
+    print(instance_en_us)
+for j in Encoding_indexesEN_IE:
+    instance_en_ie = []
+    instance_en_ie.append(capitalized[j].lower())
+    print(instance_en_ie)
+for j in Encoding_indexesEN_LATN:
+    instance_en_latn = []
+    instance_en_latn.append(capitalized[j].lower())
+    print(instance_en_latn)
+for j in Encoding_indexesGA:
```
GreihMurray, line 81: The standard Python lower() method will not always work for Gaelic as it cannot properly handle diacritic characters. There are several ways to handle it though I am only familiar with one through experience. The method I am familiar with was rather involved as I had to essentially build a miniature library in order to properly handle them

F21/annoyed-raven/main.py
```diff
@@ -0,0 +1,104 @@
...
+    print(instance_tr)
+for j in Encoding_indexesEL:
+    instance_el = []
+    instance_el.append(capitalized[j].lower())
+    print(instance_el)
+for j in Encoding_indexesZH_HANS:
+    instance_zh_hans = []
+    instance_zh_hans.append(capitalized[j].lower())
+    print(instance_zh_hans)
+for j in Encoding_indexesTH:
```
GreihMurray, line 101: Thai is similar to Chinese, please refer to prior comment

PR #108, opened 2021-10-06
code review lab

ghost:


F21/lorepap/utils.py
```diff
@@ -0,0 +1,54 @@
...
+            tmp = self.word
+            tmp = tmp.replace('\u0049', '\u0131')
+        elif language=='az':
+            tmp = self.word
+            tmp = tmp.replace('\u0049','\u0131')
+        elif language=='el':
+            if tmp[-1]=='\u03a3':
+                #self._finalSigma = True
+                tmp = tmp[:-1]+'\u03c2'
+    return tmp.lower()
```
JoeLanzi, line 41: suggestion: instead of storing in "tmp" variable, you can just return it. This will omit the "tmp" variable

PR #116, opened 2021-10-08
Nikhil

nikhilkammula:


F21/nikhilkammula/code.py
```diff
@@ -0,0 +1,23 @@
...
+    else:
+        res=word.lower()
+elif l=="el":
+    res=word.lower()
+elif l=="tr" or l=="az":
+    res=word.lower()
+else:
+    res=word
+for i in res:
+  print( i,end='')
```
Bandisaiprasad, line 23: you didn't use the test cases

F21/nikhilkammula/code.py
```diff
@@ -0,0 +1,23 @@
+l=input()
+word=input()
+res=[]
+
```
Bandisaiprasad, line 4: add comments to code for each language separation

F21/nikhilkammula/code.py
```diff
@@ -0,0 +1,23 @@
+l=input()
+word=input()
+res=[]
```
Bandisaiprasad, line 3: using array is a good idea

Bandisaiprasad (changes requested):

PR #118, opened 2021-10-10
Unit testing 

BandiRajesh:


F21/bandirajesh/myers.tsv
```diff
@@ -1,3 +1,24 @@
-{"a" : 3, "b" : 4, "c" : 5}	scalene
-{"a" : 3, "b" : 3, "c" : 1}	isosceles
-{"a" : 3, "b" : 4, "c" : 100}	error
+"{""a"" "	: 	"3, "	"""b"" "	: 	"4, "	"""c"" "	: 	"5}	scalene"
```
kscanne, line 1: some odd formatting with the quoting here, but other than that these look like good tests.

PR #186, opened 2023-02-13
Lowercase function with test cases for thofstrand

thofstrand:
Created function that lower cases a variety of languages based on input. Includes many test cases. Just run the .py file in order to run the tests.

### DIFF
--- a/S23/thofstrand/lowerCase.py
+++ b/S23/thofstrand/lowerCase.py
@@ -4,8 +4,12 @@ import pandas as pd
 
 def lowerCase(word, lang):
     word = word.strip()
-    langStart = lang[:2]
+    lang = lang.split('-')
+    langStart = lang[0]
 
+    if len(langStart) != 2:
+        return "Invalid Language"
+    
     if "en" == langStart:
         lower_word = word.lower()
         return lower_word
@@ -29,14 +33,25 @@ def lowerCase(word, lang):
         lowCase = word.lower()
         caps = ['A','E','I','O','U','Á','É','Í','Ó','Ú']
         if word[0] == 'n' or word[0] == 't':
-            if word[1] in caps and word[2].isalpha(): #must check isalpha as Õ parses as `O`,`~`
-                word_list2 = [*lowCase]
-                word_list2.insert(1,'-')
-                lower_word = combineToString(word_list2)
-                return lower_word
+            if len(word) > 2:
+                if word[1] in caps and word[2].isalpha(): #must check isalpha as Õ parses as `O`,`~`
+                    word_list2 = [*lowCase]
+                    word_list2.insert(1,'-')
+                    lower_word = combineToString(word_list2)
+                    return lower_word
+                else:
+                    word = word.lower()
+                    return word
             else:
-                word = word.lower()
-                return word
+                if word[1] in caps: #must check isalpha as Õ parses as `O`,`~`
+                    word_list2 = [*lowCase]
+                    word_list2.insert(1,'-')
+                    lower_word = combineToString(word_list2)
+                    return lower_word
+                else:
+                    word = word.lower()
+                    return word
+
             
         else:
                 word = word.lower()
@@ -60,9 +75,9 @@ def combineToString(char_list):
 
 def test_main():
 
-    test_cases = pd.read_csv('../5030/S23/thofstrand/tests.tsv',sep='\t',names=["word","lang","lower"])
+    test_cases = pd.read_csv('C:\\Users\\keerp\\Documents\\CSCI-5030\\5030\\S23\\thofstrand\\tests.tsv',sep='\t',names=["word","lang","lower"])
     for i in range(len(test_cases)):
-        #word_output = lowerCase(test_cases['word'][i],test_cases['lang'][i])
+       # word_output = lowerCase(test_cases['word'][i],test_cases['lang'][i])
         #print(word_output)
         result = lowerCase(test_cases['word'][i],test_cases['lang'][i]) == test_cases['lower'][i]
         print(result)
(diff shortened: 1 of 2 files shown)


S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,76 @@
...
+        if word[-1] == 'Σ':
+            lower_word = word.lower()
+            lower_word_list = [*lower_word]
+            lower_word_list[-1] = 'ς'
+            lower_word = combineToString(lower_word_list)
+            return lower_word
+    else:
+        return word
+    
+def combineToString(char_list):
```
mepripri, line 55: I see great use of inbuilt functions, but this combineToDefined defined functions could be used in a single line to reduce extra code.

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,76 @@
...
+            lower_word = lower_word.lower()
+            return lower_word
+        else:
+            return word
+        
+    if "ga" == langStart:
+        lowCase = word.lower()
+        caps = ['A','E','I','O','U','Á','É','Í','Ó','Ú']
+        if word[0] == 'n' or word[0] == 't':
+            if word[1] in caps and word[2].isalpha(): #must check isalpha as Õ parses as `O`,`~`
```
mepripri, line 32: I quiet didn't understand the 3 letter must be a alpha or not. I think just the 1 letter must be a n or a t and the second letter in caps.

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,76 @@
+import pandas as pd
+
+
+
+def lowerCase(word, lang):
+    word = word.strip()
+    langStart = lang[:2]
```
mepripri, line 7: We can add a condition if the user is giving false/errorful language as input.

mepripri:
The code was quite understandable, but some lines could be changed/improved. I loved the use of functions in the code. Just some more test cases are missed. Overall, it was a good piece of code, I got to learn something.

S23/thofstrand/lowerCase.py
thofstrand: Good thought! I am adding this now!

S23/thofstrand/lowerCase.py
thofstrand: Essentially one of the chars ( Õ) that was an input splits into two characters (`O`,`~`) when I am separating the string by characters. I used isalpha to check for this and make sure i am not counting that character as one of the listed ones. I do see another issue as I am not checking the length of the word, so this I can fix!

S23/thofstrand/lowerCase.py
thofstrand: This might be wise but I also feel as though the main function is already way too large. I think splitting it actually makes for fewer lines of code in the long run.

mepripri: Yes, i agree to this then.

S23/thofstrand/lowerCase.py
mepripri: This code change is fine.

mepripri:
I am fine with the changes now. Signing this off :)

mepripri (approved):

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,91 @@
+import pandas as pd
+
+
+
+def lowerCase(word, lang):
+    word = word.strip()
+    lang = lang.split('-')
+    langStart = lang[0]
+
+    if len(langStart) != 2:
```
kscanne, line 10: You should look at the BCP-47 spec.  There are (many) language codes that have length 3.

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,91 @@
...
+        return "Invalid Language"
```
kscanne, line 11: This is a place to raise an exception, not return a string error message.

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,91 @@
...
+    
+    if "en" == langStart:
```
kscanne, line 13: Why just English?  Shouldn't this be the default for many languages that have case?

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,91 @@
...
+        lower_word = word.lower()
+        return lower_word
+    
+    if "tr" == langStart or 'az' == langStart:
+        
+        if "I" in word:
+            word_list = [*word]
+            indices = [i for i, x in enumerate(word_list) if x == "I"]
```
kscanne, line 21: Simpler to use the string "replace" method vs. converting to list and back?

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,91 @@
...
+            return lower_word
+        else:
+            return word
+        
+    if "ga" == langStart:
+        lowCase = word.lower()
+        caps = ['A','E','I','O','U','Á','É','Í','Ó','Ú']
+        if word[0] == 'n' or word[0] == 't':
+            if len(word) > 2:
+                if word[1] in caps and word[2].isalpha(): #must check isalpha as Õ parses as `O`,`~`
```
kscanne, line 37: There's a cleaner way to handle this; will discuss in class.

S23/thofstrand/lowerCase.py
```diff
@@ -0,0 +1,91 @@
...
+                return word
+        
+    if "el" == langStart:
+        if word[-1] == 'Σ':
+            lower_word = word.lower()
+            lower_word_list = [*lower_word]
+            lower_word_list[-1] = 'ς'
+            lower_word = combineToString(lower_word_list)
+            return lower_word
+    else:
```
kscanne, line 67: You should have many more test cases for other languages like Spanish, French, etc. that are handled like English.  These tests will fail given the current code.

kscanne (commented):
Some additional comments I would have made on an initial review.

PR #188, opened 2023-02-16
Code to Change Words to LowerCase

mepripri:


Tanvipatel23:
new Locale(language)
new Locale(language, country)
new Locale(language, country, variant)

class JavaCharactertoLowerCaseExample1 {  
  public static void main(String[] args) {  
		
		String s2 = Normalizer.normalize(s, Normalizer.Form.NFD);
    		return s2.replaceAll("(?s)\\p{InCombiningDiacriticalMarks}", "");

     char ch1, ch2, ch3, ch4;    
      ch1 = 'TANVI';  
      ch2 = 'PATEL';  
        
      ch3 = Character.toLowerCase(ch1);  
      ch4 = Character.toLowerCase(ch2);  
      String str1 = "The lowercase for the first character '" + ch1 + "' is given as: " + ch3;  
      String str2 = "The lowercase for the second character '" + ch2 + "' is given as: " + ch4;  
  
      System.out.println( str1 );  
      System.out.println( str2 );  
      }  
}  

String myString="YAŞAT Patel";
Locale trlocale= new Locale("tr-TR");
Locale enLocale = new Locale("en_US");

Log.v("mainlist", "en source: " +myString.toLowerCase(enLocale));
Log.v("mainlist", "tr source: " +myString.toLowerCase(trlocale));

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,49 @@
+import csv
+
+def splitLang(l):
```
thofstrand, line 3: This function is great for getting the first part of the language input, but we should make `l` a little more descriptive in terms of the variable name.

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,49 @@
...
+        l=l[0]
+    return l
+
+def toLower(s,lang):
+    str=''
+    first=0
+    lang=splitLang(lang)
+    if lang=='tr' or lang=='az':
+        for i in s:
+            if ord(i)==73:
```
thofstrand, line 15: While this may work, I think the use of hard coded numbers is not ideal. Maybe if the function merely searches for 'I' instead of using ord() that would be a little clearer.

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,49 @@
...
+                continue
+            else:
+                str+=s[i].lower()
+    elif lang=='el':
+        if ord(s[-1])==931:
+            str+=s[:-1].lower()
+            str+=chr(962)
+        else:
+            str=str+s.lower()
+    elif lang=='zh' or lang=='th' or lang=='ja' or lang=='en':
```
thofstrand, line 38: We don't need to specifiy these four languages, as anything else would be just lower cased. So this could just be an else:

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,49 @@
...
+    elif lang=='ga':
+        first=0
+        for i in range(len(s)):
+            if i==0:
+                if (s[0]=='n' or s[0]=='t') and s[1] in ['A','E','I','O','U','Á','É','Í','Ó','Ú']:
+                    str+=s[0].lower()+'-'+s[1].lower()
+                    first=1
+                else:
+                    str+=s[i].lower()
+            elif i==1 and first==1:
```
thofstrand, line 28: I'm not sure what the `elif` at line 28 is doing. This may be omitted as it doesn't appear to do anything as of this moment.

S23/mepripri/tests.tsv
```diff
@@ -1,18 +1,29 @@
 HELLO	en	hello
 WORLD	en-US	world
+WORLD	eng-US	Invalid Language
```
thofstrand, line 3: `eng` would likely produce an invalid language type, but it could also just be lower cased since any other language that is not of the ones specified is just lowered.

thofstrand (commented):
Overall this code looks good and you added a decent number of test cases. There is room for improvement in terms of readability and some unnecessary lines. Well done!

S23/mepripri/ToLowerCase.py
mepripri: Sure, i will work on the variable name.

S23/mepripri/ToLowerCase.py
mepripri: Yes, that also perfect that will be a direct search.

S23/mepripri/ToLowerCase.py
mepripri: But, i also have a non-language condition there. To solve this I can add that condition in elif and make the rest as else. This will be more better.

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,49 @@
...
+                str+=s[i].lower()
+    elif lang=='el':
+        if ord(s[-1])==931:
+            str+=s[:-1].lower()
+            str+=chr(962)
+        else:
+            str=str+s.lower()
+    elif lang=='zh' or lang=='th' or lang=='ja' or lang=='en':
+        str+=s.lower()
+    else:
```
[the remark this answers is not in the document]
mepripri: This is similar to the above one I guess

S23/mepripri/tests.tsv
mepripri: This is added because this doesn't fit in the language requirement. If any language is greater than or less than 2 letters will be an invalid language.

mepripri:
I have made the changes to the code. @thofstrand, Can you please review them again

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,39 @@
+import csv
+
+def splitLang(lang):
+    lang=lang.split('-')
```
kscanne, line 4: This works, but I don't really like the change of type here from str to list.

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,39 @@
...
+    return lang[0]
+
+def toLower(s,lang):
+    str=''
+    first=0
+    lang=splitLang(lang)
+    if lang=='tr' or lang=='az':
+        for i in s:
+            if i=='I':
+                str+=chr(305)
```
kscanne, line 14: Magic number.

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,39 @@
...
+            else:
+                str+=i.lower()
+    elif lang=='ga':
+        if (s[0]=='n' or s[0]=='t') and s[1] in ['A','E','I','O','U','Á','É','Í','Ó','Ú']:
+            str+=s[0].lower()+'-'+s[1:].lower()
```
kscanne, line 19: s[0] already lowercase here?

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,39 @@
...
+        else:
+            str+=s.lower()
+    elif lang=='el':
+        if ord(s[-1])==931:
```
kscanne, line 23: Magic numbers — unreadable.

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,39 @@
...
+            str+=s[:-1].lower()
+            str+=chr(962)
+        else:
+            str=str+s.lower()
+    elif len(lang)!=2:
```
kscanne, line 28: Check out the BCP-47 spec.  Language codes can have 2 or 3 letters.

S23/mepripri/ToLowerCase.py
```diff
@@ -0,0 +1,39 @@
...
+        str+='Invalid Language'
```
kscanne, line 29: You should raise an exception here, not return an error string.

PR #189, opened 2023-02-16
code sample for lab day

akumar2709:
Sorry I won't be present for the code review lab. But this is my submission for the programming assignment.

### DIFF
--- a/S23/akumar2709/main.py
+++ b/S23/akumar2709/main.py
@@ -1,0 +1,32 @@
+import csv
+
+#function converts word to lower case and provides special rules for irish and turkish since they are not covered by the book
+def toLower(input_string : str, language: str) -> str:
+  irish_vowels = "AEIOUÁÉÍÓÚ"
+  tilda_ord = 771
+#the following condition for irish uses a conditional lamda function which looks for the following rules 
+# - include a hyphen if the first alphabet is T or N and its followed by an irish vowel. It also checks for 
+# ord 771 since that means that there is a tild on the second alphabet since that could be achieved by either a singular unicode or by 2 unicodes
+# this is the reason we only check at postion[2] since if there is a tilda at [2] it means there can't be a vowel at [1]
+
+  if(language.find("ga") + 1):
+    irish_lower = lambda input_string : (input_string[0] + "-" + input_string[1:]).lower() \
+                if (input_string[0] == "t" or input_string[0] == "n") and irish_vowels.find(input_string[1]) + 1 \
+                and ord(input_string[2]) != tilda_ord  \
+                else input_string.lower()
+
+    return irish_lower(input_string)
+
+# Turkish and Azerbaijani follow similar rules to standard covertion with just a different i
+  elif((language.find("tr") + 1) or (language.find("az") + 1)):
+    return input_string.lower().replace("i", "ı")
+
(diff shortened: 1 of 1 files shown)


S23/akumar2709/main.py
```diff
@@ -0,0 +1,32 @@
+import csv
+
+#function converts word to lower case and provides special rules for irish and turkish since they are not covered by the book
+def toLower(input_string : str, language: str) -> str:
+  irish_vowels = "AEIOUÁÉÍÓÚ"
+  tilda_ord = 771
```
kscanne, line 6: There is something more general going on here that you'll want to handle in general, not just for this one combining diacritics.  We'll discuss in class when we get into Unicode more deeply, or you can stop by office hours if you're curious.

S23/akumar2709/main.py
```diff
@@ -0,0 +1,32 @@
...
+#the following condition for irish uses a conditional lamda function which looks for the following rules 
+# - include a hyphen if the first alphabet is T or N and its followed by an irish vowel. It also checks for 
+# ord 771 since that means that there is a tild on the second alphabet since that could be achieved by either a singular unicode or by 2 unicodes
+# this is the reason we only check at postion[2] since if there is a tilda at [2] it means there can't be a vowel at [1]
+
+  if(language.find("ga") + 1):
```
kscanne, line 12: != -1 would be more readable.  Also, this isn't exactly correct, since there are 3-letter language codes in ISO 639-3, and so this would work for "gaa", for example (Ga language spoken in Ghana).

S23/akumar2709/main.py
```diff
@@ -0,0 +1,32 @@
...
+    irish_lower = lambda input_string : (input_string[0] + "-" + input_string[1:]).lower() \
```
kscanne, line 13: A bit odd to do this as a lambda.  Why not just apply these lines of code directly to the object with the same name?

S23/akumar2709/main.py
```diff
@@ -0,0 +1,32 @@
...
+                if (input_string[0] == "t" or input_string[0] == "n") and irish_vowels.find(input_string[1]) + 1 \
+                and ord(input_string[2]) != tilda_ord  \
+                else input_string.lower()
+
+    return irish_lower(input_string)
+
+# Turkish and Azerbaijani follow similar rules to standard covertion with just a different i
+  elif((language.find("tr") + 1) or (language.find("az") + 1)):
```
kscanne, line 21: Same issue with 3-letter codes.

S23/akumar2709/main.py
```diff
@@ -0,0 +1,32 @@
...
+    return input_string.lower().replace("i", "ı")
```
kscanne, line 22: I don't believe this behaves the way we want.  Did you test with both dotted and non-dotted uppercase?

kscanne: You *really* need a lot more tests, since they might turn up additional bugs.

S23/akumar2709/main.py
```diff
@@ -0,0 +1,32 @@
...
+
+  else:
+    return input_string.lower()
```
kscanne, line 25: The problem description asked for an optimization for languages without case.

PR #191, opened 2023-02-16
Java lowercase function

fulvius31:
Created a sample application in java + gradle for lowercase task assigned. I've also created a junit test so it can be tested.

Since I don't have the permission to write any files outside my folder I've created my own repo so I can use github actions to make automatic tests based on `tests.tsv` file (https://github.com/fulvius31/5030-workflow). I've also added some other test cases in `tests.tsv` file.

# How to run the test cases manually :

- clone the repo
- select the branch `java_lowercase`
- import the folder `fulvius31` in IntelliJ IDE or an IDE that supports Gradle
- open `app/src/test/java/lowercase/AppTest.java`
- press the `play` button placed on the top right bar


S23/fulvius31/app/src/main/java/lowercase/Lowercaser.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private boolean isExceptionLang() {
+        return lang.equals(LanguagesToExamine.TURKISH.getAcronym()) ||
+                lang.equals(LanguagesToExamine.AZERBAIJANI.getAcronym()) ||
+                lang.equals(LanguagesToExamine.IRISH.getAcronym()) ||
+                lang.equals(LanguagesToExamine.GREEK.getAcronym());
+    }
+
+    enum LanguagesToExamine {
```
PMIRUKULA, line 95: Maybe we can use any function that would work for any language instead of dynamically giving some languages to the code (Locale)

fulvius31: Thanks for your review!

Could you provide me some examples and why it would be better than use an enum ?

S23/fulvius31/app/src/main/java/lowercase/App.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * This Java source file was generated by the Gradle 'init' task.
+ */
+package lowercase;
+
+public class App {
+
+    public static void main(String[] args) {
+        Lowercaser lowercaser = new Lowercaser("nÕg","ga-US");
```
PMIRUKULA, line 9: Instead of changing input in the code every time.can we just read it using scanner/buffer reader.

S23/fulvius31/app/src/main/java/lowercase/Lowercaser.java
PMIRUKULA: by enum data structure you have only hard coded few of the languages but dynamically we can't use enum. It would be better we we give languages dynamically

fulvius31: thanks for your comment on it. 

enum is not hard coded, enum is a data type. It is used - in this case - also to improve the readability and the maintainability of the code. For example, if you have to add "it" language, maybe the reader does not know that "it" is for the italian language. So who writes the code, knows it and put it in an enum as ITALIAN with its acronym "it".

The purpose of the function was not to make the application interactive with the users but to test it thru JUnit tests (in this case).

Tell me if you need more details.

S23/fulvius31/app/src/main/java/lowercase/App.java
fulvius31: I've just removed the useless code on main.

S23/fulvius31/.gitattributes
```diff
@@ -0,0 +1,9 @@
+#
```
kscanne, line 1: Slightly painful to have to navigate through all these accessory files to get to the actual lowercasing code.  Sometimes nice to do a PR with just the reviewable code.

S23/fulvius31/app/src/main/java/lowercase/Lowercaser.java
```diff
@@ -0,0 +1,115 @@
+package lowercase;
+
+import java.text.Normalizer;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+public class Lowercaser {
+
+    private final String upperCase;
```
kscanne, line 10: Slightly unusual design here; I might have guessed that a "Lowercaser" class represented just the "lowercasing algorithm", in which case you'd only want to pass the language code to the constructor, and then pass the string you want to lowercase as an argument to a "lowercase" method.

S23/fulvius31/app/src/main/java/lowercase/Lowercaser.java
```diff
@@ -0,0 +1,115 @@
...
+                strChar[i] = Character.toLowerCase(strChar[i]);
+        }
+        return new String(strChar);
+    }
+
+    private String lowerGa() {
+        Set<String> irishVowels = new HashSet<>
+                (Arrays.asList("A", "E", "I", "O", "U", "Á", "É", "Í", "Ó", "Ú"));
+        String normalizedUpper = Normalizer.normalize(upperCase,
+                Normalizer.Form.NFC);
```
kscanne, line 33: Good!

S23/fulvius31/app/src/main/java/lowercase/Lowercaser.java
```diff
@@ -0,0 +1,115 @@
...
+            return lowerTrAz();
+        }
+        if (lang.equals(LanguagesToExamine.IRISH.getAcronym())) {
+            return lowerGa();
+        }
+        if (lang.equals(LanguagesToExamine.GREEK.getAcronym())) {
+            return lowerEl();
+        }
+
+        return upperCase;
```
kscanne, line 79: Does this ever get called?

S23/fulvius31/app/src/main/java/lowercase/Lowercaser.java
```diff
@@ -0,0 +1,115 @@
...
+        for (int i = 0; i < strCharLength; i++) {
+            strChar[i] = Character.toLowerCase(strChar[i]);
+        }
+        return new String(strChar);
+    }
+
+    public String getLowerCase() {
+        if (isUnchangedLang())
+            return upperCase;
+        if (!isExceptionLang())
```
kscanne, line 65: Maybe make this the else case; this would avoid having to add new exceptional languages in two places (easy to forget).

PR #192, opened 2023-02-16
Add files via upload

PMIRUKULA:
On running my code, a text file would be generated in same location with output in it.


S23/pmirukula/Conversion.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.io.*;
+import java.util.Locale;
+
+public class Conversion {
+	public void conversionlogic() 
+	{
+	}
+
```
fulvius31, line 11: // Is this function used somewhere ? Do we need it?

S23/pmirukula/Conversion.java
```diff
@@ -0,0 +1,44 @@
...
+	public static void main(String[] args) throws IOException {
+		BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+		System.out.println("Please enter the word");
+		String str = br.readLine();
+		System.out.println("Please enter the language you would like to translate the word");
+		String str_lang = br.readLine();
+		PrintWriter out = new PrintWriter("oceans.txt");
```
fulvius31, line 18: It would be better to use a better name for the file to improve the readability

S23/pmirukula/Conversion.java
```diff
@@ -0,0 +1,44 @@
...
+
+	public static void main(String[] args) throws IOException {
+		BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+		System.out.println("Please enter the word");
+		String str = br.readLine();
+		System.out.println("Please enter the language you would like to translate the word");
+		String str_lang = br.readLine();
+		PrintWriter out = new PrintWriter("oceans.txt"); 
+		String output = str.toLowerCase(new Locale(str_lang));
+		if(str_lang == " tr " || str_lang == "az")
```
fulvius31, line 20: // we should use the appropriate method for string called equals. More explaination here https://www.geeksforgeeks.org/difference-between-and-equals-method-in-java/
		// there is a white space in " tr". It was made on purpose?

S23/pmirukula/Conversion.java
```diff
@@ -0,0 +1,44 @@
...
+		{
+			if(str.contains("I")) 
+			{
+				str = str.replaceAll("I", " \0u131");
+				output = str.toLowerCase(new Locale(str_lang));
+			}
+		}
+		else if(str_lang.equals("ga-IE")|| str_lang.equals("ga") )
```
fulvius31, line 28: // maybe it could be better to handle the string in a way there we can separate two part of the language. It could be awesome if we can split the string with "-" sign so we can just compare it with "ga".

S23/pmirukula/Conversion.java
```diff
@@ -0,0 +1,44 @@
...
+		{
+	
+			if((str.charAt(0) == 'n' || str.charAt(0) == 't')&& (str.charAt(1) == 'A' || str.charAt(1) == 'E' || str.charAt(1) == 'I' || str.charAt(1) == 'O' 
+			|| str.charAt(1) == 'U'|| str.charAt(1) == 'Ó' || str.charAt(1) == 'Á' ||str.charAt(1) == 'É' || str.charAt(1) == 'Ú' ||str.charAt(1) == 'Í'))
```
fulvius31, line 32: // It could be better for readability to have a shorter statement

fulvius31 (changes requested):
@PMIRUKULA  It should be nice if you can provide me a way to test the code.

Did you try all the test cases ?

S23/pmirukula/Conversion.java
PMIRUKULA: I had originally intended to write a method rather than writing the entire thing in the main, but I was short on time. I'll discontinue using the technique.
Thank you

S23/pmirukula/Conversion.java
PMIRUKULA: Noted! will work on it.
Thanks for Suggestion!

S23/pmirukula/Conversion.java
PMIRUKULA: No, it was a mistake. will change it!
Thank you.

PMIRUKULA:
> @PMIRUKULA It should be nice if you can provide me a way to test the code.
> 
> Did you try all the test cases ?

yes, I have tested all the cases.
You can just compile it in java compiler. read the output you want and then a file oceans.txt file would be generated with output in it.

fulvius31:
Maybe it would be better to use JUnit tests so I can easily watch on them and test it.

S23/pmirukula/Conversion.java
kscanne: Writing this as a main function only is not an ideal design choice, since the function cannot then be used as library function in other code.  I'd refactor into its own re-usable class.

S23/pmirukula/Conversion.java
```diff
@@ -0,0 +1,44 @@
...
+		String str = br.readLine();
+		System.out.println("Please enter the language you would like to translate the word");
+		String str_lang = br.readLine();
+		PrintWriter out = new PrintWriter("oceans.txt"); 
+		String output = str.toLowerCase(new Locale(str_lang));
+		if(str_lang == " tr " || str_lang == "az")
+		{
+			if(str.contains("I")) 
+			{
+				str = str.replaceAll("I", " \0u131");
```
kscanne, line 24: Did you test this?  Looks like a bug.  And also a magic number.

S23/pmirukula/Conversion.java
kscanne, line 28: BCP-47 codes can be very complicated. This is far from general enough.

S23/pmirukula/Conversion.java
kscanne: You shouldn't resolve these comments without actually making the changes on your branch.

PR #194, opened 2023-02-16
Convert to LowerCase ( code review lab)

MVR9999:
AllTestCasesToLowerCase.java file has all the test cases. ConvertToLowerCase.java, we give input and expect the output result.txt has all the tested test cases


S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,100 @@
+package com.firstspringproject.Controllers;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
+import java.util.*;
+
+public class ToLowerCase {
+    
+    String turkish_lang = "tr";
```
Sandeep-1-Kumar, line 9: I feel that naming conventions for attribute names in camelCase looks better rather than separating with "_"
Ex:  turkish_lang to turkishLang

S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,100 @@
...
+            // TODO Auto-generated catch block
+            e.printStackTrace();
+        }
+        ToLowerCase object = new ToLowerCase();
+      for(int i=0;i<words.length;i++)
+      {
+        String result = object.wordToLowerCase(words[i],lang[i],langWithNoLowerCase);
+        printWriter.println(words[i]+ " " + lang[i] +" "+result);
+      }
+
```
Sandeep-1-Kumar, line 55: I can suggest that elimination of unused extra empty lines, It might make code looks messy.

S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,100 @@
...
+        {
+            if(word.contains(upper_Case_I))
+            {
+                word.replaceAll(upper_Case_I,latin_Dotless_I );
+                lower1 = word.toLowerCase(locale1);
+            }
+        }
+        else if(language.startsWith(irish_lang))
+        {
+            if((word.charAt(0) == lowerCase_n || word.charAt(0) == lowerCase_t) &&
```
Sandeep-1-Kumar, line 77: A better logic for this implementation is to use for loop , defining the possibilities in an array. As you can see that if else multiple conditions is not an good way of sorting.

S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,100 @@
...
+                word.charAt(1) == upperCase_Irish_O || word.charAt(1) == upperCase_Irish_U))
+            {
+                lower1 = lower1.substring(0,1) +"-" + lower1.substring(1);
+            }
+        }
+      
+            return lower1;
+    }
+
+    boolean isLanguage_with_no_lowerCase(String []langWithNoLowerCase,String language)
```
Sandeep-1-Kumar, line 91: I felt that in the above method , method name looks good than here as it uses camel casing. You can code better following single naming convection for all the methods in the class.

S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,100 @@
+package com.firstspringproject.Controllers;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
+import java.util.*;
+
+public class ToLowerCase {
```
Sandeep-1-Kumar, line 7: I think in java a class name should be the file name they should match. However we know that we can just have a single class in single .java file , Its a good way to name the class after file.

S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,100 @@
...
+    int upperCase_Irish_O = 211;
+    int upperCase_Irish_U = 218;
+    public static void main(String args[]) throws UnsupportedEncodingException
+    {
+        // words to be converted to lowerCase
+        String words[] = {"HELLO", "WORLD", "cAmEl", "---OK", "tAcht", "tACHT", "TACHT", "nAthair", "nATHAIR",
+        "NATHAIR", "tOthair", "tETHAIR", "tITHAIR", "nÓg", "nÕg","对不起","ごめんなさい", "KASIM", "KASIM", "ΠΌΛΗΣ", "官话", "ภาษาไทย", "ΠΌΛΗΣ", "ΠΌΛΗΣΠΌΛΗΣ"};
+        
+        // words in their respective languages 
+        String lang[] = {"en", "en-US", "en-IE", "en-Latn", "ga", "ga", "ga", "ga", "ga",
```
Sandeep-1-Kumar, line 37: I think when one value in an array satisfies condition duplicates are not necessary in array -> "ga".

S23/mvr9999/AllTestCasesToLowerCase.java
MVR9999: noted!

S23/mvr9999/AllTestCasesToLowerCase.java
MVR9999: noted! we shall implement this better using a for loop

MVR9999:
Hello, I have modified the code as per the given suggestions and added the JUnit test cases, Could you please confirm

Sandeep-1-Kumar:
Sign Off all verified.

S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.*;
+
+public class AllTestCasesToLowerCase {
+    
+    String turkishLang = "tr";
+    String azerbaijaniLang = "az";
+    String irishLang = "ga";
+    String latinDotlessI = "\u0131";
+
+    String upperCaseI = "I";
```
kscanne, line 14: These three maybe unnecessary?  I'm happy with a literal 't' for example.

S23/mvr9999/ConvertToLowerCase.java
```diff
@@ -0,0 +1,88 @@
+package com.firstspringproject.Controllers;
+
+import java.util.*;
+
+public class ConvertToLowerCase {
```
kscanne, line 5: I don't like the repeated code here and in AllTestCasesToLower....

S23/mvr9999/AllTestCasesToLowerCase.java
```diff
@@ -0,0 +1,103 @@
...
+                if(word.charAt(1) == vowel)
+                {
+                lower1 = lower1.substring(0,1) +"-" + lower1.substring(1);
+                break;
+                }
+
+            } }
+        }
+      
+            return lower1;
```
kscanne, line 91: No Greek?

PR #195, opened 2023-02-16
vivekupstreamAssignment

vivekkatari:


S23/vivekkatari/Main.java
```diff
@@ -0,0 +1,45 @@
+import java.util.Locale;
+import java.util.Scanner;
+
+public class Main {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.print("Enter a word: ");
+        String word = scanner.nextLine();
```
Tanvipatel23, line 8: It would be easy to write a word in a new line and  it can be easy to find a word .

S23/vivekkatari/Main.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public static class LowerCase {
+        public static String toLowerCase(String word, String language) {
+            Locale locale = language != null && !language.isEmpty() ? Locale.forLanguageTag(language) : Locale.getDefault();
+
+            String lowerCaseWord;
+            String languageTag = locale.toLanguageTag();
+
+            if (languageTag.startsWith("tr") || languageTag.startsWith("az")) lowerCaseWord = word.replace("I", "ı");
```
Tanvipatel23, line 23: I would be nice if you use brackets .

Tanvipatel23 (changes requested):

S23/vivekkatari/Main.java
vivekkatari: Thanks for the suggestion, but i feel there wont be difference if we write code in different lines

S23/vivekkatari/Main.java
vivekkatari: Thanks for the suggestion The functionality is work as expected so there would be no difference in updating the code.

PR #196, opened 2023-02-16
Lowercaseit

austin-carnahan:


b-sai (commented):
I like how the script runs without using for loops, makes it readable and clear. However, more test cases are not included as per directions on canvas. Another minor point is that the lowercasing for ga is supposed to work for any variant of ga. Ditto for all other languages. But otherwise great stuff!

austin-carnahan:
> I like how the script runs without using for loops, makes it readable and clear. However, more test cases are not included as per directions on canvas. Another minor point is that the lowercasing for ga is supposed to work for any variant of ga. Ditto for all other languages. But otherwise great stuff!

austin-carnahan:
Thanks for the review! I fixed the parts of the code that were missing use cases - especially the Irish languages.

PR #197, opened 2023-02-16
Add files via upload

kishore223:


S23/kishore223/lowercase.py
```diff
@@ -0,0 +1,24 @@
...
+            if word[i-1] in ['n', 't'] and word[i] in vowels and word[i].isupper():
+                return word[:i] + '-' + word[i:].lower()
+        return word.lower()
+    elif lang == 'el':
+        if word.endswith('Σ'):
+            return word[:-1] + 'ς'
+        else:
+            return word.lower()
+    elif lang in ['zh', 'ja', 'th']:
+        return word
```
pbutiya, line 16: Here in the line 16th we can write return word.lower()

S23/kishore223/lowercase.py
```diff
@@ -0,0 +1,24 @@
...
+    else:
+        return word.lower()
+
+word = input("Enter a word: ")
+lang = input("Enter language code (BCP-47): ")
+
+lowercased_word = lowercase_word(word, lang)
```
pbutiya, line 23: if some person will enter any random BCP-47 code which is not existing then it will convert text into lower case but after the output coming how person can know that the language code is correct or not so should should we ask for language inspite of language code?

S23/kishore223/lowercase.py
kishore223: Chinese (zh), Japanese (ja), or Thai (th) these languages don't have notion of upper and lowercase characters. So it returns unchanged.

S23/kishore223/lowercase.py
kishore223: Now I have mentioned the code word for every language.

S23/kishore223/lowercase.py
```diff
@@ -0,0 +1,80 @@
+import unittest
+
+def lowercase_word(word, lang):
+    if lang in ['tr', 'az']:
+        return word.replace('I', 'ı').lower()
+    elif lang == 'ga':
+        vowels = set(['A', 'E', 'I', 'O', 'U', 'Á', 'É', 'Í', 'Ó', 'Ú'])
+        for i in range(1, len(word)):
+            if word[i-1] in ['n', 't'] and word[i] in vowels and word[i].isupper():
+                return word[:i] + '-' + word[i:].lower()
```
kscanne, line 10: This should only be applied at the beginning of a word.    nAthair -> n-athair, but if it's "bhurnAthair, then "bhurnathair" is fine.

S23/kishore223/lowercase.py
```diff
@@ -0,0 +1,80 @@
+import unittest
+
+def lowercase_word(word, lang):
+    if lang in ['tr', 'az']:
```
kscanne, line 4: If "lang" is meant to be a BCP-47 code, you'll need to pick out the language code from the beginning.

